PR #1, opened 2018-01-16
Fix a minor 'go vet' error in httpcommon.go

KenjiKanedaNVIDIA:
This commit will fix the following error by adding 'defer cancel()'

```
httpcommon.go:144: the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak
```

C.f.) https://golang.org/pkg/context/#WithTimeout

### DIFF
--- a/dfc/httpcommon.go
+++ b/dfc/httpcommon.go
@@ -141,7 +141,8 @@ func (r *httprunner) run() error {
 func (r *httprunner) stop(err error) {
 	glog.Infof("Stopping %s, err: %v", r.name, err)
 
-	contextwith, _ := context.WithTimeout(context.Background(), ctx.config.HttpTimeout)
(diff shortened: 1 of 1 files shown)


dfc/httpcommon.go
```diff
@@ -141,7 +141,8 @@ func (r *httprunner) run() error {
 func (r *httprunner) stop(err error) {
 	glog.Infof("Stopping %s, err: %v", r.name, err)
 
-	contextwith, _ := context.WithTimeout(context.Background(), ctx.config.HttpTimeout)
+	contextwith, cancel := context.WithTimeout(context.Background(), ctx.config.HttpTimeout)
+	defer cancel()
```
ambarsarkar: Using defer cancel is recommended best practice.

alex-aizman:
done quite some time ago

PR #4, opened 2018-04-07
diskkeeper update

VladimirMarkelov:
1. Renamed to FSKeeper since it works with filesystems
2. Added a config option to disable FSKeeper
3. Make temporary file name follow rules for temporary DFC file
(isworkfile must return true for the name)

### DIFF
--- a/dfc/fskeeper.go
+++ b/dfc/fskeeper.go
@@ -8,7 +8,7 @@ package dfc
 import (
 	"io/ioutil"
 	"os"
-	"sort"
+	"path"
 	"sync/atomic"
 	"time"
 
@@ -16,10 +16,12 @@ import (
 )
 
 const (
-	fsCheckInterval = time.Second * 10
+	fsCheckInterval     = time.Second * 30
+	tmpDirnameTemplate  = "DFC-TEMP-DIR"
+	tmpFilenameTemplate = "DFC-TEMP-FILE"
 )
 
-type diskkeeper struct {
+type fskeeper struct {
 	namedrunner
 	checknow chan error
 	chstop   chan struct{}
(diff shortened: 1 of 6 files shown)


dfc/fskeeper.go
```diff
@@ -29,8 +31,8 @@ type diskkeeper struct {
 }
 
 // construction
-func newdiskkeeper(t *targetrunner) *diskkeeper {
-	k := &diskkeeper{t: t}
+func newfskeeper(t *targetrunner) *fskeeper {
+	k := &fskeeper{t: t}
```
usmanong: return &fskeeper{t:t} without the k?

dfc/fskeeper.go
```diff
@@ -39,31 +41,30 @@ func newdiskkeeper(t *targetrunner) *diskkeeper {
...
-	k.okmap.okmap[sid] = time.Now()
+	k.okmap.okmap[mpath] = time.Now()
 	k.okmap.Unlock()
 }
 
-func (k *diskkeeper) skipCheck(sid string) bool {
+func (k *fskeeper) skipCheck(mpath string) bool {
 	k.okmap.Lock()
-	last, ok := k.okmap.okmap[sid]
+	last, ok := k.okmap.okmap[mpath]
```
usmanong: can return here if not ok

dfc/target.go
```diff
@@ -578,22 +575,23 @@ func (t *targetrunner) listCachedObjects(bucket string, msg *GetMsg) (outbytes [
...
 		glog.Errorf(errstr)
+		return
 	}
 
+	var reslist = BucketList{Entries: allfinfos.files}
+	// Mark the batch as truncated if it is full
+	if allfinfos.fileCount >= allfinfos.limit {
+		reslist.PageMarker = allfinfos.lastFilePath
+	}
+	outbytes, err = json.Marshal(reslist)
```
usmanong: return can be removed

PR #6, opened 2018-04-30
Authn

usmanong:


authn/authn_test.go
```diff
@@ -0,0 +1,204 @@
+package main
+
+import (
```
usmanong: Glad to see we have unit tests

dfc/proxy.go
```diff
@@ -1870,3 +1895,84 @@ func (p *proxyrunner) broadcast(urlfmt, method string, jsbytes []byte, callback
...
+	}
+
+	newAuth, err := newAuthList(tokenList)
+	if err != nil {
+		s := fmt.Sprintf("Invalid token list: %v", err)
+		p.invalmsghdlr(w, r, s)
+		return
+	}
+
+	p.authList = newAuth
```
usmanong: lock for authList also needed I believe.

PR #9, opened 2018-05-09
Add target logic for registering/unregistering self

liangdrew:
A message signal to register/unregister a target is passed from the target API once it is hit from the proxy side. The register/unregister signal is consumed in the keepalive runner to start/stop the ticker that re-registers a target (or proxy) runner.


dfc/proxy.go
```diff
@@ -1839,18 +1845,21 @@ func (p *proxyrunner) httpcludel(w http.ResponseWriter, r *http.Request) {
...
 	p.smap.lock()
 	defer p.smap.unlock()
-	if !proxy && p.smap.get(sid) == nil {
+	si := p.smap.get(sid)
+	if !proxy && si == nil {
 		glog.Errorf("Unknown target %s", sid)
 		return
 	}
-	if proxy && p.smap.getProxy(sid) == nil {
+	if proxy && si == nil {
```
VladimirMarkelov: smap methods get/del/add are a bit misleading. They work only with Tmap part, while getProxy/delProxy/addProxy work only with Pmap part. So, the patch does invalid check here.

We may think of adding 'Target' to 'add/del/etc' to avoid confusion in the future.

dfc/proxy.go
```diff
@@ -1756,6 +1758,10 @@ func (p *proxyrunner) httpclupost(w http.ResponseWriter, r *http.Request) {
 	} else {
 		newtarget = &nsi
 		p.registertarget(&nsi, keepalive)
+		if register {
```
usmanong: if it fails?

dfc/proxy.go
```diff
@@ -1839,18 +1845,21 @@ func (p *proxyrunner) httpcludel(w http.ResponseWriter, r *http.Request) {
 	}
 	p.smap.lock()
 	defer p.smap.unlock()
-	if !proxy && p.smap.get(sid) == nil {
+	si := p.smap.get(sid)
+	if !proxy && si == nil {
```
usmanong: duplicate code? these three lines and the next?

PR #10, opened 2018-05-10
Add stress test

liangdrew:
1. Unregister target T
2. Create local bucket
3. PUT large amount of objects into the local bucket
4. Randomly GET the objects while simultaneously re-registering the target T

### DIFF
--- a/dfc/tests/regression_test.go
+++ b/dfc/tests/regression_test.go
@@ -20,6 +20,8 @@ import (
 	"testing"
 	"time"
 
+	"sync/atomic"
+
 	"github.com/NVIDIA/dfcpub/dfc"
 	"github.com/NVIDIA/dfcpub/pkg/client"
 	"github.com/NVIDIA/dfcpub/pkg/client/readers"
@@ -28,6 +30,11 @@ import (
 type Test struct {
 	name   string
 	method func(*testing.T)
+}
+
+type repFile struct {
+	repetitions int
+	filename    string
 }
 
 type regressionTestData struct {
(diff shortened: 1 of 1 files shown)


dfc/tests/regression_test.go
```diff
@@ -88,7 +93,115 @@ var (
...
+				repFile := <-repFilenameCh
+				if repFile.repetitions > 0 {
+					repFile.repetitions -= 1
+					repFilenameCh <- repFile
+				}
+				_, _, err := client.Get(proxyurl, bucket, SmokeStr+"/"+repFile.filename, nil, nil, false, false)
+				if err != nil {
+					numGetErrs += 1
+				}
+				wg.Done()
```
usmanong: this is ok, but good practice is use defer wg.done() at the top of the function in case there are returns in the middle

dfc/tests/smoke_test.go
```diff
@@ -218,7 +218,9 @@ func fillWithRandomData(seed int64, fileSize uint64, objList []string, bucket st
 			}
 			errch <- err
 		}
-		filesput <- fname
+		if filesput != nil {
```
usmanong: are there any places that calls this with files put == nil? if not, no need to check, it will crash if a nil is passed here and coder will find that out right away, there is no runtime possible calling this with ch == nil

dfc/tests/regression_test.go
```diff
@@ -88,7 +93,115 @@ var (
...
+// 2. Create local bucket
+// 3. PUT large amount of objects into the local bucket
+// 4. GET the objects while simultaneously re-registering the target T
+func TestGetAndReRegisterInParallel(t *testing.T) {
+	var (
+		sid                string
+		num                = 20000
+		numGetsForEachFile = 5
+		numGetErrs         int
+		maxNumGetErrs      = num * numGetsForEachFile / 10000 // 0.01% of GET requests
```
usmanong: 0.01% is not 10,000, this is 10%
add an explanation of why less than this % is expected to fail.

liangdrew: Yup, forgot to change the percentage in the comment. Thanks for the catch!

PR #12, opened 2018-05-10
Remove target out of metasyncer

usmanong:


usmanong:
@alex-aizman @VladimirMarkelov @liangdrew 
mostly shuffle code around, but the key change is take target out of metasyncer, code is simpler and precise of who does what and better functions ownership. 
thanks

PR #13, opened 2018-05-10
Rename local bucket sequence change

usmanong:
1. Only tell targets to do rename
2. Sync lbmap to all targets and proxy

### DIFF
--- a/dfc/proxy.go
+++ b/dfc/proxy.go
@@ -837,6 +837,8 @@ func (p *proxyrunner) renamelocalbucket(bucketFrom, bucketTo string, lbmap4ren *
 	msg.Value = lbmap4ren
 	jsbytes, err := json.Marshal(msg)
 	assert(err == nil, err)
+	// only needs to tell target of the rename, proxies will get the sync call later
+	smap4bcast.Pmap = nil
 	p.broadcast(urlfmt, method, jsbytes, smap4bcast, f, ctx.config.Timeout.Default)
 
 	lbmapLock.Lock()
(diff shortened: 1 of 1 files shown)


usmanong:
@alex-aizman @sasanap @VladimirMarkelov @liangdrew

dfc/proxy.go
```diff
@@ -849,6 +851,8 @@ func (p *proxyrunner) renamelocalbucket(bucketFrom, bucketTo string, lbmap4ren *
 		glog.Errorf("Failed to store lbmap %s, err: %v", lbpathname, err)
```
VladimirMarkelov: It is not a question to current changes: should we get back old bucket name to lbmap in case of LocalSave fails. 
1) I remember rollback is implemented on save error in many places of metasync patch but not there
2) it seems, now if save fails, a client receives an error and the client tries again, but error changes to 'bucket does not exist' because lbmap is changed?

usmanong: agree, we need a proper rollback, not just here, may be other places. I will add a comment for now and may be work on some of these rollbacks in a separate change
thanks

dfc/proxy.go
```diff
@@ -837,6 +837,8 @@ func (p *proxyrunner) renamelocalbucket(bucketFrom, bucketTo string, lbmap4ren *
 	msg.Value = lbmap4ren
 	jsbytes, err := json.Marshal(msg)
 	assert(err == nil, err)
+	// only needs to tell target of the rename, proxies will get the sync call later
```
liangdrew: Nitpick: consider moving these two lines right after `p.smap.copyL(smap4bcast)` so all the modifications to `smap4bcast` are in the same place.

usmanong: will do
thanks

PR #14, opened 2018-05-11
DFCLoader failed to reuse already existing bucket

usmanong:


usmanong:
@alex-aizman @VladimirMarkelov @liangdrew @sasanap 
this fixes Sachin's dfcloader problem
thanks

PR #18, opened 2018-05-12
Resolve webdav flag name collision

liangdrew:
webdav is panicking at runtime due to flag redefinition


webdav/server.go
```diff
@@ -43,14 +43,14 @@ func main() {
 	)
 
 	flag.IntVar(&port, "port", 8079, "this server's port")
-	flag.StringVar(&proxy, "proxyurl", "127.0.0.1:8080", "dfc proxy's url (ip:port)")
-	flag.StringVar(&tmpDir, "tmpdir", "/tmp/dfc", "temporary directory to store files used by webdav")
-	flag.IntVar(&logLevel, "loglevel", logLevelNone, "verbose level(0 = none, 1 = webdav, 2 = dfc)")
+	flag.StringVar(&proxy, "proxyURL", "127.0.0.1:8080", "dfc proxy's URL (ip:port)")
```
usmanong: this is suppose to be proxy url, this is the command line option's shown name, why change to proxyURL? same for tmpDir and LogLevel

webdav/server.go
```diff
@@ -43,14 +43,14 @@ func main() {
...
+	flag.StringVar(&tmpDir, "tmpDir", "/tmp/dfc", "temporary directory to store files used by webdav")
+	flag.IntVar(&logLevel, "logLevel", logLevelNone, "verbose level(0 = none, 1 = webdav, 2 = dfc)")
 	flag.Parse()
-	url := url.URL{Scheme: "http", Host: proxy}
+	dfcURL := url.URL{Scheme: "http", Host: proxy}
```
usmanong: url is very much locally used and golfing encourage's short local names, making it dfcURL doesn't add much value.

usmanong:
if those three flags (proxyurl, log level and tmpdir) are conflicting with ddc's init(), i think they can be removed from here as long as they have the same meaning)

usmanong:
i think it is the case, likely one of my changes that made daemon.go's init function create some conflicts, proxy url and log level.
the proper fix is remove proxy url from webdav, reuse daemon's
log level in wendav can be renamed to something, it is a sort of temporary work around fro webdav development, as it is stable for now, it can be set to just verbose or depends on the default option for glog -v

webdav/server.go
liangdrew: The variable name `url` here conflicts with the import package with the same name.

usmanong: really? that is strange, it is a local variable, conflict with which one?

PR #20, opened 2018-05-14
Primary proxy change

VladimirMarkelov:
We had two places to keep the current primary proxy. Most of the time they contained the same value. But it was confusing and it was hard to get what and where to change. The patch leaves only one place to keep primary proxy. With the patch proxy test suit passes more test than the code with 2 places: 8/10 against 5/10. Only putGet & stress are failing. 
The proxy test is still disabled.

### DIFF
--- a/dfc/httpcommon.go
+++ b/dfc/httpcommon.go
@@ -138,7 +138,6 @@ type httprunner struct {
 	httpclientLongTimeout *http.Client // http client for long-wait intra-cluster comm
 	statsif               statsif
 	kalive                kaliveif
-	proxysi               *proxyInfo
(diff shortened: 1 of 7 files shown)


VladimirMarkelov:
1. It can be nil. At least I can name two places: on proxy startup if it is primary (if it is non-primary then it receives smap and it should contains proxySI), and on target startup - it seems proxysi is nil unti the target connected to proxy successfully.

2. Yes, it is still a dup. It can be replaced with utility function that traverse the proxy array and return one that has primary: true

usmanong:
go ahead merge it

PR #22, opened 2018-05-15
Unit tests for meta sync data

usmanong:
this is the second part of adding unit test for meta sync code, it helps me to confirm what the code suppose to do
I have another one to do which is check delete and the sync map
thanks

### DIFF
--- a/dfc/metasync_test.go
+++ b/dfc/metasync_test.go
@@ -582,3 +582,90 @@ func TestMetaSyncData(t *testing.T) {
 	syncer.stop(nil)
 	wg.Wait()
 }
+
+// TestMetaSyncMembership tests meta sync's logic when accessing proxy's smap directly
+func TestMetaSyncMembership(t *testing.T) {
+	{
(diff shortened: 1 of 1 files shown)


dfc/metasync_test.go
```diff
@@ -362,15 +373,213 @@ func refused(t *testing.T, primary *proxyrunner, syncer *metasyncer) ([]transpor
...
+	for i := 0; i < cnt; i++ {
+		act := (<-ch).data
+		for k, e := range act {
+			a, ok := exp[k]
+			if !ok {
+				fail(t, exp, act)
+			}
+
+			if strings.Compare(e, a) != 0 {
+				fmt.Println("xxx ", k, e, a)
```
VladimirMarkelov: Debug?

usmanong: true, removed, thanks

usmanong: diff updated

PR #23, opened 2018-05-16
Unit tests for meta sync membership

usmanong:
last one, with this one, metasync.go is 95.6% covered, only lines not covered are log lines and data validation which will be removed later.
thanks


usmanong:
thank you

PR #25, opened 2018-05-16
Add test for GET after file corruption in cache

liangdrew:
~Also adding object FQN as a response header to the HEAD object API (so that it can be used to corrupt the test files).~ Traversing the directory tree to find FQNs.

### DIFF
--- a/dfc/tests/regression_test.go
+++ b/dfc/tests/regression_test.go
@@ -22,6 +22,8 @@ import (
 
 	"sync/atomic"
 
+	"path/filepath"
+
 	"github.com/NVIDIA/dfcpub/dfc"
 	"github.com/NVIDIA/dfcpub/pkg/client"
 	"github.com/NVIDIA/dfcpub/pkg/client/readers"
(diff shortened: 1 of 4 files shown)


pkg/client/client.go
```diff
@@ -285,12 +286,13 @@ func get(proxyurl, bucket string, keyname string, wg *sync.WaitGroup, errch chan
 
 	v := hdhashtype == dfc.ChecksumXXHash
 	len, hash, err := readResponse(resp, w, err, fmt.Sprintf("GET (object %s from bucket %s)", keyname, bucket), v)
-	if err != nil && v {
+	if v {
```
usmanong: good catch, hopefully it doesn't triggers new test errors that were hidden before

VladimirMarkelov:
What about corruption and PUT? Is it the next patch?

Does test pass? I do not see in out code checking hash when reading.

pkg/client/client.go
```diff
@@ -285,12 +286,13 @@ func get(proxyurl, bucket string, keyname string, wg *sync.WaitGroup, errch chan
...
 		if hdhash != hash {
 			s := fmt.Sprintf("Header's hash %s doesn't match the file's %s \n", hdhash, hash)
 			if errch != nil {
```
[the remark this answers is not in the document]
liangdrew: yup thanks, good catch

dfc/tests/regression_test.go
```diff
@@ -95,6 +99,74 @@ var (
...
+		}
+		return nil
+	})
+	if errstr := dfc.Setxattr(fqn, dfc.XattrXXHashVal, []byte("corrupted xattr")); errstr != "" {
+		t.Error(errstr)
+	}
+	_, _, err = client.Get(proxyurl, bucket, SmokeStr+"/"+f, nil, nil, false, true)
+	if err == nil {
+		t.Error("Err is nil, expected non-nil error on a GET for an object with corrupted xattr")
+	}
```
VladimirMarkelov: files created by putRandomFiles should be deleted at the end of the test. As an example function 'deletefromfilelist' in main_test

PR #27, opened 2018-05-16
Meta syncer avoid creating tickers

usmanong:
@alex-aizman @VladimirMarkelov @liangdrew @sasanap 

thanks

### DIFF
--- a/dfc/metasync.go
+++ b/dfc/metasync.go
@@ -102,7 +102,7 @@ type metasyncer struct {
 	chfeed     chan []interface{}
 	chfeedwait chan []interface{}
 	chstop     chan struct{}
-	ticker     *time.Ticker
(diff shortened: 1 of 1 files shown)


dfc/metasync.go
```diff
@@ -116,7 +116,7 @@ func newmetasyncer(p *proxyrunner) (y *metasyncer) {
 	y.chfeedwait = make(chan []interface{})
 
 	// create the ticker, do not need to start at start up
```
liangdrew: Seems like this comment is misleading. Remove?

usmanong: will change it
will also change the ticker name, i was debating about it, think of it now, 'timer' is a better name, and it is actually a timer not a ticker
thank you both for the review

PR #28, opened 2018-05-17
Misc metasync code changes

usmanong:
1. Consolidate pending check in one place.
2. Modified pending handler and refused handler logic by using channel instead of lock. Using lock has a race between
   the actual for loop and the callbacks.
3. Added a small delay in the refused test case, a small % of the test didn't go to that route, this delay
   helps to make it happen as the test case is designed for.
4. Fixed a few locking issue in test code


usmanong:
not a lot code change but lots logic changes, with unit tests in place, it allows me to make quick changes with confidence.

Please review.
Thanks

PR #29, opened 2018-05-17
Modify async target re-reg and GET test

liangdrew:
Now, split the GET errors into two categories:

1) GET errors occurring before the target has updated its local bucket map
2) GET errors occurring after the target has updated its local bucket map

### DIFF
--- a/dfc/tests/regression_test.go
+++ b/dfc/tests/regression_test.go
@@ -147,7 +147,7 @@ func TestGetCorruptFileAfterPut(t *testing.T) {
 	}
 	_, _, err = client.Get(proxyurl, bucket, SmokeStr+"/"+f, nil, nil, false, true)
 	if err == nil {
-		t.Error("Nil error received on a GET for a corrupted object")
+		t.Error("Error is nil, expected non-nil error on a a GET for an object with corrupted contents")
 	}
 
 	// Test corrupting the file xattr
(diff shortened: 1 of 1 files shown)


dfc/tests/regression_test.go
```diff
@@ -260,38 +263,56 @@ func TestGetAndReRegisterInParallel(t *testing.T) {
 				}
 				_, _, err := client.Get(proxyurl, bucket, SmokeStr+"/"+repFile.filename, nil, nil, false, false)
 				if err != nil {
-					atomic.AddUint64(&numGetErrs, 1)
+					if reregistered == 1 {
```
VladimirMarkelov: reregistered is set using atomics but read without it

dfc/tests/regression_test.go
```diff
@@ -260,38 +263,58 @@ func TestGetAndReRegisterInParallel(t *testing.T) {
...
+					if r == 1 {
+						atomic.AddUint64(&numGetErrsAfter, 1)
+					} else {
+						atomic.AddUint64(&numGetErrsBefore, 1)
+					}
 				}
 				atomic.AddUint64(&getsCompleted, 1)
-				if (getsCompleted)%5000 == 0 {
-					tlogf(" %d/%d GET requests completed...\n", getsCompleted, num*numGetsForEachFile)
+				g := atomic.LoadUint64(&getsCompleted)
```
VladimirMarkelov: nitpick: AddUint64 returns a new value, so you can combine two lines in one

liangdrew: I used a variable here because the value is used twice below

VladimirMarkelov: I meant this line and above, not below )

PR #31, opened 2018-05-17
HTTP API for set glog's vmodule

usmanong:

thank you

### DIFF
--- a/dfc/config.go
+++ b/dfc/config.go
@@ -344,3 +344,20 @@ func setloglevel(loglevel string) (err error) {
 	}
 	return
 }
+
+// setGLogVModule sets glog's vmodule flag
+// sets 'v' as is, no verificaton is done here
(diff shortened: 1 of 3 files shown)


dfc/config.go
```diff
@@ -344,3 +344,20 @@ func setloglevel(loglevel string) (err error) {
 	}
 	return
 }
+
+// setGLogVMoudle sets glog's vmodule flag
+// sets 'v' as is, no verificaton is done here
+// syntax for v: target=5,proxy=1, p*=3, etc
+func setGLogVMoudle(v string) error {
```
liangdrew: `setGLogVMoudle` -> `setGLogVModule`

usmanong: of course
i wonder when will nvidia's AI can do spelling check on source code, but if that day do come, probably we all will lose our jobs
thanks for catching it

PR #32, opened 2018-05-18
Refactor and add integration test

liangdrew:
The sequence for the new test TestProxyFailbackAndGetAndReRegisterInParallel is as follows:

1. Unregister a target
2. Create a local bucket
3. Crash the primary proxy and PUT in parallel
4. Failback to the original primary proxy, re-register target, and GET in parallel

### DIFF
--- a/dfc/tests/integration_test.go
+++ b/dfc/tests/integration_test.go
@@ -1,0 +1,326 @@
+package dfc_test
+
+import (
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"runtime/debug"
+
+	"github.com/NVIDIA/dfcpub/dfc"
+	"github.com/NVIDIA/dfcpub/pkg/client"
+)
+
+type repFile struct {
+	repetitions int
+	filename    string
+}
+
+type metadata struct {
+	t                   *testing.T
+	smap                dfc.Smap
+	delay               time.Duration
+	semaphore           chan struct{}
+	controlCh           chan struct{}
+	repFilenameCh       chan repFile
+	wg                  *sync.WaitGroup
+	bucket              string
+	targetDirectURL     string
+	sid                 string
+	otherTasksToTrigger int
+	origNumTargets      int
+	num                 int
+	numGetsEachFile     int
+	numGetErrsBefore    uint64
+	numGetErrsAfter     uint64
+	getsCompleted       uint64
+	reregistered        uint64
+}
+
+// Intended for a deployment with multiple targets
+// 1. Unregister target T
+// 2. Create local bucket
+// 3. PUT large amount of objects into the local bucket
+// 4. GET the objects while simultaneously re-registering the target T
+func TestGetAndReRegisterInParallel(t *testing.T) {
+	const (
+		num      = 20000
+		filesize = uint64(1024)
+		seed     = int64(111)
+	)
+
+	var (
+		err error
+		m   = metadata{
+			t:               t,
+			delay:           18 * time.Second,
+			num:             num,
+			numGetsEachFile: 5,
+			repFilenameCh:   make(chan repFile, num),
+			semaphore:       make(chan struct{}, 10), // 10 concurrent GET requests at a time
+			wg:              &sync.WaitGroup{},
+			bucket:          TestLocalBucketName,
+		}
+		// Currently, a small percentage of GET errors can be reasonably expected as a result of this test.
+		// With the current design of dfc, there is exists a brief period in which the cluster map is synced to
+		// all nodes in the cluster during re-registering. During this period, errors can occur.
+		filenameCh    = make(chan string, m.num)
+		errch         = make(chan error, 100)
+		sgl           *dfc.SGLIO
+		maxNumGetErrs = uint64(m.num * m.numGetsEachFile / 10) // 10 % of GET requests
+	)
+	// Step 1.
+	m.smap, err = client.GetClusterMap(proxyurl)
+	checkFatal(err, t)
+
+	m.origNumTargets = len(m.smap.Tmap)
+	if m.origNumTargets < 2 {
+		t.Fatalf("Must have 2 or more targets in the cluster, have only %d", m.origNumTargets)
+	}
+	for m.sid = range m.smap.Tmap {
+		break
+	}
+	m.targetDirectURL = m.smap.Tmap[m.sid].DirectURL
+	err = client.UnregisterTarget(proxyurl, m.sid)
+	checkFatal(err, t)
+
+	n := len(getClusterMap(httpclient, t).Tmap)
+	if n != m.origNumTargets-1 {
+		t.Fatalf("%d targets expected after unregister, actually %d targets", m.origNumTargets-1, n)
+	}
+	tlogf("Unregistered target %s: the cluster now has %d targets\n", m.sid, n)
+
+	// Step 2.
+	err = client.CreateLocalBucket(proxyurl, m.bucket)
+	checkFatal(err, t)
+
+	defer func() {
+		err = client.DestroyLocalBucket(proxyurl, m.bucket)
+		checkFatal(err, t)
+	}()
+
+	if usingSG {
+		sgl = dfc.NewSGLIO(filesize)
+		defer sgl.Free()
+	}
+
+	// Step 3.
+	tlogf("Putting %d files into bucket %s...\n", num, m.bucket)
+	putRandomFiles(0, seed, filesize, num, m.bucket, t, nil, errch, filenameCh, SmokeDir, SmokeStr, "", true, sgl)
+	selectErr(errch, "put", t, false)
+	close(filenameCh)
+
+	for f := range filenameCh {
+		m.repFilenameCh <- repFile{repetitions: m.numGetsEachFile, filename: f}
+	}
+
+	// Step 4.
+	m.wg.Add(num*m.numGetsEachFile + 1)
+	go doReregisterTarget(&m)
+	doGetsInParallel(&m)
+
+	m.wg.Wait()
+	tlogf("%d GET errors before target's local bucket map was updated (expected)\n", m.numGetErrsBefore)
+	if m.numGetErrsAfter > maxNumGetErrs {
+		t.Fatalf("Found %d GET errors (should be no more than %d/%d) after target's local bucket map was updated",
+			m.numGetErrsAfter, maxNumGetErrs, num*m.numGetsEachFile)
+	} else {
+		tlogf("Found %d GET errors (should be no more than %d/%d) after target's local bucket map was updated",
+			m.numGetErrsAfter, maxNumGetErrs, num*m.numGetsEachFile)
+	}
+}
+
+// 1. Unregister a target
+// 2. Create a local bucket
+// 3. Crash the primary proxy and PUT in parallel
+// 4. Failback to the original primary proxy, re-register target, and GET in parallel
+func TestProxyFailbackAndGetAndReRegisterInParallel(t *testing.T) {
+	const (
+		num                 = 20000
+		otherTasksToTrigger = 1
+		filesize            = uint64(1024)
+		seed                = int64(111)
+	)
+
+	var (
+		err error
+		m   = metadata{
+			t:                   t,
+			delay:               15 * time.Second,
+			otherTasksToTrigger: otherTasksToTrigger,
+			num:                 num,
+			numGetsEachFile:     5,
+			repFilenameCh:       make(chan repFile, num),
+			semaphore:           make(chan struct{}, 10), // 10 concurrent GET requests at a time
+			controlCh:           make(chan struct{}, otherTasksToTrigger-1),
+			wg:                  &sync.WaitGroup{},
+			bucket:              TestLocalBucketName,
+		}
+		// Currently, a small percentage of GET errors can be reasonably expected as a result of this test.
+		// With the current design of dfc, there is exists a brief period in which the cluster map is synced to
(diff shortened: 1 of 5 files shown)


dfc/tests/objprops_test.go
```diff
@@ -223,7 +223,8 @@ func propsRebalance(t *testing.T, bucket string, objects map[string]string, msg
 	}
 
 	tlogf("Removing a target: %s\n", removedSid)
-	unregisterTarget(removedSid, t)
+	err := client.UnregisterTarget(proxyurl, removedSid)
+	checkFatal(err, t)
 	waitProgressBar("Removing: ", time.Second*10)
```
usmanong: this wait vs the wait in UnregisterTarget(), both are necessary?

dfc/tests/integration_test.go
```diff
@@ -0,0 +1,301 @@
...
+	}()
+
+	if usingSG {
+		sgl = dfc.NewSGLIO(filesize)
+		defer sgl.Free()
+	}
+
+	// Step 3.
+	tlogf("Putting %d files into bucket %s...\n", num, m.bucket)
+	putRandomFiles(0, seed, filesize, num, m.bucket, t, nil, errch, filenameCh, SmokeDir, SmokeStr, "", true, sgl)
```
usmanong: use m.num?

dfc/tests/integration_test.go
```diff
@@ -0,0 +1,301 @@
...
+	"github.com/NVIDIA/dfcpub/dfc"
+	"github.com/NVIDIA/dfcpub/pkg/client"
+)
+
+type repFile struct {
+	repetitions int
+	filename    string
+}
+
+type metadata struct {
```
usmanong: in general, reduce scope of variables are better, when i see a big struct like this, i will see if they are all necessary to put into one struct, some can be local.
i did see 'm' is passed into two calls, even if they share most of the fields, still can be separated into sub struct.
i am ok with this as this is a test code, just some thoughts.

dfc/tests/integration_test.go
```diff
@@ -0,0 +1,301 @@
...
+
+// Intended for a deployment with multiple targets
+// 1. Unregister target T
+// 2. Create local bucket
+// 3. PUT large amount of objects into the local bucket
+// 4. GET the objects while simultaneously re-registering the target T
+func TestGetAndReRegisterInParallel(t *testing.T) {
+	var (
+		err error
+		num = 20000
```
usmanong: make this a const

dfc/tests/integration_test.go
```diff
@@ -0,0 +1,301 @@
...
+		errch         = make(chan error, 100)
+		sgl           *dfc.SGLIO
+		filesize      = uint64(1024)
+		seed          = int64(111)
+	)
+	// Step 1.
+	m.smap, err = client.GetClusterMap(proxyurl)
+	checkFatal(err, t)
+
+	l := len(m.smap.Tmap)
```
usmanong: i would do:
m.origNumTragets = len(m.smap.Tmap)
if m.origNumTargets < 2 {
}

to avoid using 'l', unless your concern is too verbose if using m.origNumTargets three times
i am ok if you keep it this way

dfc/tests/integration_test.go
```diff
@@ -0,0 +1,301 @@
...
+	for f := range filenameCh {
+		m.repFilenameCh <- repFile{repetitions: m.numGetsEachFile, filename: f}
+	}
+	m.wg.Wait()
+
+	// Step 4
+	m.wg.Add(m.num*m.numGetsEachFile + 2)
+	go func() {
+		// Failback is timed to ensure it doesn't finish before GETs begin
+		time.Sleep(17 * time.Second)
```
VladimirMarkelov: Interesting number of seconds. Can we avoid sleeps with random or hand-picked on certain machine number of seconds? As an idea: what if we share a channel between this function and doRepeatedGets(maybe we can do something for the sleep above). This function tries to read from channel and hangs until doRepeatedGets closes the channel and then continues. doRepeated can do a few GETs, then close channel and continue doing GETs, so you get two "sub-tests" reading a little % of data from stable cluster and the other reads go to the cluster while it is changing its config

dfc/tests/integration_test.go
```diff
@@ -0,0 +1,301 @@
...
+	}
+	m.origNumTargets = l
+	for m.sid = range m.smap.Tmap {
+		break
+	}
+	m.targetDirectURL = m.smap.Tmap[m.sid].DirectURL
+	err = client.UnregisterTarget(proxyurl, m.sid)
+	checkFatal(err, t)
+
+	n := len(getClusterMap(httpclient, t).Tmap)
```
usmanong: is this needed? should it be just m.origNumTargets -1?
better use of this call may be verify only one target went away after unregister

dfc/tests/integration_test.go
```diff
@@ -0,0 +1,301 @@
...
+	tlogf("Unregistered target %s: the cluster now has %d targets\n", m.sid, n)
+
+	// Step 2.
+	m.bucket = TestLocalBucketName
```
usmanong: move this to where m is created as other fields do?

dfc/tests/integration_test.go
```diff
@@ -0,0 +1,301 @@
...
+	}
+
+	// Step 4.
+	m.wg.Add(num*m.numGetsEachFile + 1)
+	go doRepeatedGets(&m)
+	doReregisterTarget(&m)
+
+	m.wg.Wait()
+	tlogf("%d GET errors before target's local bucket map was updated (expected)\n", m.numGetErrsBefore)
+	if m.numGetErrsAfter > maxNumGetErrs {
```
VladimirMarkelov: You calculate m.numGetErrsBefore as well. But I do not see where the total number is used.

dfc/tests/integration_test.go
```diff
@@ -0,0 +1,301 @@
...
+	selectErr(errch, "put", t, false)
+	close(filenameCh)
+
+	for f := range filenameCh {
+		m.repFilenameCh <- repFile{repetitions: m.numGetsEachFile, filename: f}
+	}
+
+	// Step 4.
+	m.wg.Add(num*m.numGetsEachFile + 1)
+	go doRepeatedGets(&m)
```
usmanong: wg.Add is better to go with the actual coroutines are started,  in the for loop.
actually doRepeatedGets() doesn't have to be run in a go routine, if you do, it is actually not waiting for the period you want to wait inside doRepeatedGets(), it didn't wait and started the register right away below

dfc/tests/integration_test.go
```diff
@@ -0,0 +1,301 @@
...
+}
+
+func doReregisterTarget(m *metadata) {
+	err := client.RegisterTarget(m.sid, &(m.smap))
+	checkFatal(err, m.t)
+PollLoop:
+	for i := 0; i < 25; i++ {
+		time.Sleep(time.Second)
+		m.smap = getClusterMap(httpclient, m.t)
+		lbNames, err := client.GetLocalBucketNames(m.targetDirectURL)
```
VladimirMarkelov: Maybe move this line into the next `if`? It does not makes sense to get bucket names from a target if it is not in smap yet.

dfc/tests/integration_test.go
```diff
@@ -0,0 +1,301 @@
...
+	if len(m.smap.Tmap) != m.origNumTargets {
+		m.t.Fatalf("Re-registration timed out: target %s, num targets now: %d, orig num targets: %d\n",
+			m.sid, len(m.smap.Tmap), m.origNumTargets)
+	}
+	tlogf("Re-registered target %s: the cluster is now back to %d targets\n", m.sid, m.origNumTargets)
+	s := atomic.CompareAndSwapUint64(&m.reregistered, 0, 1)
+	if !s {
+		m.t.Errorf("reregistered should have swapped from 0 to 1. Actual reregistered = %d\n", m.reregistered)
+	}
+	m.wg.Done()
```
VladimirMarkelov: I'd move it to the beginning of the function as deferred call

dfc/tests/integration_test.go
```diff
@@ -0,0 +1,301 @@
...
+}
+
+func checkFatal(err error, t *testing.T) {
```
VladimirMarkelov: I am not sure. On the one hand, it saves 2 lines every time. From the other hand, it hides the place of a fail - you always see in logs: "Failed at line 299" that is kind of useless. For small tests it is OK, for big ones it is not that good. 
Another way: it is possible to get the callers info(it is done in DFC) and tlogf the line which raises the error(httpcommon.go:Write)

dfc/tests/regression_test.go
```diff
@@ -813,7 +650,7 @@ func regressionRebalance(t *testing.T) {
 	for sid = range smap.Tmap {
 		break
 	}
-	unregisterTarget(sid, t)
+	client.UnregisterTarget(proxyurl, sid)
```
VladimirMarkelov: returned error is not checked

pkg/client/client.go
```diff
@@ -992,6 +992,18 @@ func GetPrimaryProxy(url string) (string, error) {
 	return smap.ProxySI.DirectURL, nil
 }
 
+func UnregisterTarget(url, sid string) error {
+	err := HTTPRequest("DELETE", url+"/"+dfc.Rversion+"/"+dfc.Rcluster+"/"+"daemon"+"/"+sid, nil)
+	// FIXME: do something better than sleep
+	time.Sleep(time.Second * 3)
```
VladimirMarkelov: I do not think it should be in client. In my opinion it should better be in test where it calls the client library

pkg/client/client.go
```diff
@@ -992,6 +992,18 @@ func GetPrimaryProxy(url string) (string, error) {
...
+	return err
+}
+
+func RegisterTarget(sid string, smap *dfc.Smap) error {
```
VladimirMarkelov: Since the function does not check anything, it can be turned into one-liner:
return HTTPRequest("POST", smap.Tmap[sid].DirectURL+"/"+dfc.Rversion+"/"+dfc.Rdaemon+"/"+dfc.Rregister, nil)
In this case, it is not clear why we need it in the library

dfc/tests/integration_test.go
liangdrew: `m.numGetErrsBefore` is logged, but not checked because they are expected failures. Discussed this with @usmanong earlier today.

dfc/tests/integration_test.go
liangdrew: `num` is defined at the top of the function as well, so `num` and `m.num` are the same here.

dfc/tests/integration_test.go
VladimirMarkelov: Sorry. I called Ctlf+F but did not notice that it is logged.

pkg/client/client.go
VladimirMarkelov: While driving home, I thought about this way to replace waiting with random time. It seems sleep is done to be sure that all nodes have updated Smap. Maybe replace Sleep with a function that reads Smap from alive proxies and targets and compares their Smap version? And it loops until timeout or versions are the same

pkg/client/client.go
```diff
@@ -1075,3 +1096,40 @@ func DoesLocalBucketExist(serverURL string, bucket string) (bool, error) {
 
 	return false, nil
 }
+
+func WaitMapVersionSync(timeout time.Time, smap dfc.Smap, idsToIgnore []string) error {
+	prev := smap.Version
+	urls := make(map[string]struct{})
+	for _, d := range smap.Tmap {
+		for _, ignore := range idsToIgnore {
```
VladimirMarkelov: when I was fed up with writing similar loops I added a simple function to regression_test: stringInSlice :) 
It returns bool, so you can replace it with if !stringInSlice(d.DaemonID, idsToIgnore) { urls[....
But if you keep loops, it is fine to me

pkg/client/client.go
```diff
@@ -1075,3 +1096,40 @@ func DoesLocalBucketExist(serverURL string, bucket string) (bool, error) {
 
 	return false, nil
 }
+
+func WaitMapVersionSync(timeout time.Time, smap dfc.Smap, idsToIgnore []string) error {
```
VladimirMarkelov: If I get it right, the function does not wait anything. It checks the list of urls only once and sleeps between every urls. I thougth it would check all urls without sleeping in between and there would be outer loop that stops when url list is empty with sleep after inner loop finishes.

pkg/client/client.go
liangdrew: Thanks for the heads up! I think I'll keep this loop as-is for now (since `stringInSlice()` is in a different package), but the next time I have to do something like this, I'll add a generic function to `utils.go` (or somewhere that makes sense).

pkg/client/client.go
liangdrew: There's a `continue` that will short-circuit each sleep on a successful check. The loop breaks when the map is empty.

pkg/client/client.go
VladimirMarkelov: Yes, I agree. I missed that it is in different packages. But util.go function may be helpful next time.

pkg/client/client.go
VladimirMarkelov: continue goes to the next item. So, when the loop reached end of the list it stops. What restarts the loop if anything is left in the list?

liangdrew: The loop iterates over the map `urls`. With each successful check, it removes the corresponding url from the map. Sorry, I think I'm a little confused by your comment. Is this what you are asking about?

PR #34, opened 2018-05-18
Add option to get target URL on list bucket

liangdrew:
Also add local and cloud bucket tests and update the README.

~An updated README (for this advanced usage) will come after.~ Also, PUT requests sent directly to the target (i.e. not from a proxy) will be disallowed in a following patch to prevent abuse.

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,16 +235,16 @@ More usage examples can be found in the [the source](dfc/tests/regression_test.g
 
 ## List Bucket
 
-the ListBucket API returns a page of object names (and, optionally, their properties including sizes, creation times, checksums, and more), in addition to a token allowing the next page to be retrieved.
+The ListBucket API returns a page of object names (and, optionally, their properties including sizes, creation times, checksums, and more), in addition to a token allowing the next page to be retrieved.
 
 ### properties-and-options
 The properties-and-options specifier must be a JSON-encoded structure, for instance '{"props": "size"}' (see examples). An empty structure '{}' results in getting just the names of the objects (from the specified bucket) with no other metadata.
 
 | Property/Option | Description | Value |
 | --- | --- | --- |
-| props | The properties to return with object names | A comma-separated string containing any combination of: "checksum","size","atime","ctime","iscached","bucket","version". <sup id="a6">[6](#ft6)</sup> |
+| props | The properties to return with object names | A comma-separated string containing any combination of: "checksum","size","atime","ctime","iscached","bucket","version","targetUrl". <sup id="a6">[6](#ft6)</sup> |
 | time_format | The standard by which times should be formatted | Any of the following [golang time constants](http://golang.org/pkg/time/#pkg-constants): RFC822, Stamp, StampMilli, RFC822Z, RFC1123, RFC1123Z, RFC3339. The default is RFC822. |
-| prefix | The prefix which all returned objects must have. | For example, "my/directory/structure/" |
+| prefix | The prefix which all returned objects must have | For example, "my/directory/structure/" |
 | pagemarker | The token identifying the next page to retrieve | Returned in the "nextpage" field from a call to ListBucket that does not retrieve all keys. When the last key is retrieved, NextPage will be the empty string |
 | pagesize | The maximum number of object names returned in response | Default value is 1000. GCP and local bucket support greater page sizes. AWS is unable to return more than [1000 objects in one page](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketGET.html). |\b
 
--- a/dfc/REST.go
+++ b/dfc/REST.go
@@ -149,6 +149,7 @@ const (
 	GetPropsIsCached = "iscached"
 	GetPropsBucket   = "bucket"
 	GetPropsVersion  = "version"
+	GetTargetURL     = "targetUrl"
 )
 
 //===================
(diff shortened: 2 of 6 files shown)


dfc/REST.go
```diff
@@ -72,13 +72,14 @@ const (
...
-	GetPageMarker string `json:"pagemarker"`  // AWS/GCP: marker
-	GetPageSize   int    `json:"pagesize"`    // maximum number of entries returned by list bucket call
+	GetWhat       string `json:"what"`           // "config" | "stats" ...
+	GetSort       string `json:"sort"`           // "ascending, atime" | "descending, name"
+	GetProps      string `json:"props"`          // e.g. "checksum, size" | "atime, size" | "ctime, iscached" | "bucket, size"
+	GetTimeFormat string `json:"time_format"`    // "RFC822" default - see the enum below
+	GetPrefix     string `json:"prefix"`         // object name filter: return only objects which name starts with prefix
+	GetPageMarker string `json:"pagemarker"`     // AWS/GCP: marker
+	GetPageSize   int    `json:"pagesize"`       // maximum number of entries returned by list bucket call
+	GetTargetURL  bool   `json:"get_target_url"` // if should return the target URL for each object on list bucket
```
VladimirMarkelov: nitpick: I think 'get' should be removed from json:get_target_url. All others do not start from get

dfc/gcp.go
```diff
@@ -194,9 +194,16 @@ func (gcpimpl *gcpimpl) listbucket(ct context.Context, bucket string, msg *GetMs
 
 		reslist.Entries = append(reslist.Entries, entry)
 	}
+
 	if glog.V(4) {
 		glog.Infof("listbucket count %d", len(reslist.Entries))
 	}
+
+	if msg.GetTargetURL {
```
VladimirMarkelov: I do not think this way works. gcp & aws modules does not know where their objects are. When proxy gets a request for cloud bucket list, it asks a random target for it. So, the returned result is incorrect.
One way to fix: ask every target about their objects and add DirectURL depending on the information. It may not work well if objects are not cached yet.
Another way: get the list from one target and then at proxy side traverse the list and using hrw* function detect which target is responsible. Then add its DirectURL

usmanong: or is this case is only needed for local buckets?

dfc/REST.go
liangdrew: how's `return_target_url`?

dfc/gcp.go
VladimirMarkelov: As I understand the idea, we are going to provide direct link to targets (not depending on if object is in cloud or local bucket). In all cases the responsible target is one selected by hrw*. If you ask incorrect target to get an object, in best case you get performance drop because the target will go to cloud for the object - it should not cache it because it does not belong to its cache.

README.md
```diff
@@ -244,9 +244,10 @@ The properties-and-options specifier must be a JSON-encoded structure, for insta
...
-| prefix | The prefix which all returned objects must have. | For example, "my/directory/structure/" |
+| prefix | The prefix which all returned objects must have | For example, "my/directory/structure/" |
 | pagemarker | The token identifying the next page to retrieve | Returned in the "nextpage" field from a call to ListBucket that does not retrieve all keys. When the last key is retrieved, NextPage will be the empty string |
-| pagesize | The maximum number of object names returned in response | Default value is 1000. GCP and local bucket support greater page sizes. AWS is unable to return more than [1000 objects in one page](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketGET.html). |\b
+| pagesize | The maximum number of object names returned in response | Default value is 1000. GCP and local bucket support greater page sizes. AWS is unable to return more than [1000 objects in one page](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketGET.html). |
+| get_target_url | A boolean representing whether to return the URL of the target responsible for each object in the response | Default is `false`. |\b
```
VladimirMarkelov: remove trailing 'get'

dfc/gcp.go
VladimirMarkelov: For local buckets this code and the code in aws.go are never executed

usmanong: does aws or google cloud even have a concept of which 'host/target'? is it always go to s3.aws.com?

VladimirMarkelov: What about cached data? I thought the typical use-case of s3 with directurl is:  they do prefetch, so data go from cloud to local targets and cached. Then they call list with directUrls. It would be as fast as local bucket and no cloud interaction.

Otherwise I do not see sense in providing direct url for s3/aws - if they call s3/aws instead of DFC then why do they need DFC at all? :)

dfc/REST.go
liangdrew: I want to be explicit here that the property **isn't** a string representing the actual target's URL, rather, a boolean representing whether a target URL should be returned.

liangdrew:
Fixed setting the target URL for cloud buckets - and the numbers went down too. Went from ~18s for 100 objects to ~12s for 100 objects (for the new cloud bucket test).

dfc/REST.go
VladimirMarkelov: By the way, why did you add a new property to request instead of adding a new property to return? I expected that you add GetPropsTargetURL and send it as a member of GetProps field.
TargetURL is a new field of object property list returned by listbucket. It looks strange to have all other fields in GetProps but extract TaregtURL prop to a separate bool in request body

dfc/tests/regression_test.go
```diff
@@ -92,6 +92,119 @@ var (
 	failLRU = ""
 )
 
+// Note: This test is most effective when multiple targets are deployed.
```
VladimirMarkelov: You can add check for targetCount > 1 in the test and generate t.Warning if targets == 1 :) So, people will be informed that the test is not reliable

dfc/tests/regression_test.go
```diff
@@ -92,6 +92,119 @@ var (
...
+		l, _, err := client.Get(e.TargetURL, bucket, e.Name, nil, nil, false, false)
+		checkFatal(err, t)
+		if uint64(l) != filesize {
+			t.Errorf("Expected filesize: %d, actual filesize: %d\n", filesize, l)
+		}
+	}
+
+	smap, err := client.GetClusterMap(proxyurl)
+	checkFatal(err, t)
+	if len(smap.Tmap) != len(targets) { // The objects should have been distributed to all targets
```
VladimirMarkelov: It seems that was why you asked about even distribution.
Not sure if this check should be here. If you generate TargetURL incorrectly then direct object request to the URL should fail because rebalance is not running and target is not able to find the target inside data. 
Moreover, if you change test later to have fewer files it may fail even if everything else is good (random is random).

VladimirMarkelov: It applies to local buckets only. For cloud ones I do not know a reliable way to check - direct URL request should work even if you try to download object from invalid target

dfc/tests/regression_test.go
```diff
@@ -92,6 +92,119 @@ var (
...
+		t.Errorf("Expected %d different target URLs, actual: %d different target URLs", len(smap.Tmap), len(targets))
+	}
+}
+
+// Note: This test is most effective when multiple targets are deployed.
+func TestCloudListBucketGetTargetURL(t *testing.T) {
+	const (
+		num      = 20
```
VladimirMarkelov: It is again about the last check in the function. If you run many targets it is probable that your 20 objects will be on less targets than the test expects.
Maybe make check less strict? E.g, check that the number of different targets in response in greater than 1(for case number of targets > 1) or 1(for 1 target)?

dfc/tests/regression_test.go
```diff
@@ -92,6 +92,119 @@ var (
...
+
+	err := client.CreateLocalBucket(proxyurl, bucket)
+	checkFatal(err, t)
+
+	defer func() {
+		err = client.DestroyLocalBucket(proxyurl, bucket)
+		checkFatal(err, t)
+	}()
+
+	putRandomFiles(0, seed, filesize, num, bucket, t, nil, errch, filenameCh, SmokeDir, SmokeStr, "", true, sgl)
```
VladimirMarkelov: If every put fails the test is stuck: error channel length is only 100

dfc/tests/regression_test.go
liangdrew: This is a good point, but I think it's a good idea to sanity check this here. Besides, the cluster map is fetched anyway for emitting a warning when only 1 target is deployed. Also, I think it's fairly safe to trust the randomness with these numbers (according to Li's earlier stats).

VladimirMarkelov: OK. Since you increased the number of objects to put, it looks good to me now

dfc/REST.go
liangdrew: Slacked ur bro

PR #35, opened 2018-05-19
Refactor broadcast and call code

usmanong:

much cleaner code for broadcasting.
thanks

### DIFF
--- a/dfc/httpcommon.go
+++ b/dfc/httpcommon.go
@@ -16,12 +16,15 @@ import (
 	"log"
 	"net"
 	"net/http"
+	"net/url"
 	"os"
+	"path"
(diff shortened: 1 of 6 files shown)


dfc/httpcommon.go
```diff
@@ -757,3 +760,37 @@ func (h *httprunner) extractlbmap(payload map[string]string) (newlbmap *lbmap, m
 	}
 	return
 }
+
+// URLPath returns a HTTP URK path by joining all segments with "/"
```
liangdrew: `URK` -> `URL`

usmanong: done, thanks

PR #37, opened 2018-05-21
Allows detecting server timeout/failure using different/configurable …

usmanong:
…stratergy

1. code cleanup
2. fix a bug or may be unintentional behavior
   current code works in term of keeping target and primary proxy alive and communicate each
   other, but i doubt it is the intention, here is how current code works:
   a. target send keepalive every 20s if there is no communication with primary for the past 20s
   b. primary receives keepalive from target, all good
   c. next target keepalive come in 20.000009s, primary thinks it is late(by .000009), so it asks
      target for a health check.
   d. target receives the health check, mark down the time last communicated with primary
      primary also records down the time it has communicated with the target triggered by the health
      check which is a register call
   e. 20s passed, target suppose to send keepalive, but it found out that it has communicated with
      primary just recently, so it skips the keepalive
   f. primary keepalive found out there is no keepalive, health check again
   so the actual keep alive between pimary and target is like kept alive by the sequence of:
   keepalive, health check, health check, keepalive, etc, random sequence
   i fixed it by letting target always send keepalive on the fixed time, no skipping, also record
   last communicate time with primary after a successful keepalive, this makes the keepalive sequence
   between primary and target as:
   keepalive, keepalive, keepalive, ... no health check at all unless abnormal things happen
3. new interface KeepaliveTracker to allow plugable failure detection
4. new average keepalive interval tracker(not used yet, requires config file change,
   will do in follow up change)
5. unit test cases for the two keepalive tracker implementation

Thanks

### DIFF
--- a/dfc/keepalivetracker.go
+++ b/dfc/keepalivetracker.go
@@ -1,0 +1,129 @@
+/*
+ * Copyright (c) 2018, NVIDIA CORPORATION. All rights reserved.
+ */
+
+package dfc
+
+// This file has different implementations of KeepaliveTracker.
+// ideally this file stays under dfcpub, but currently dfc has a flat structure, following that
+// for now.
+import (
+	"time"
+)
+
+// HeartBeatTracker tracks timestamp of the last a message is received from a server.
+// Timeout: a message is not received within the interval.
+type HeartBeatTracker struct {
+	ch       chan struct{}
+	last     map[string]time.Time
+	interval time.Duration // expected to hear from the server within the interval
+}
+
+// NewHeartBeatTracker returns a HeartBeatTracker
+func NewHeartBeatTracker(interval time.Duration) *HeartBeatTracker {
+	hb := &HeartBeatTracker{last: make(map[string]time.Time), ch: make(chan struct{}, 1)}
+	hb.interval = interval
+	hb.unlock()
+	return hb
+}
+
+func (hb *HeartBeatTracker) lock() {
+	<-hb.ch
+}
+
+func (hb *HeartBeatTracker) unlock() {
+	hb.ch <- struct{}{}
+}
+
+// HeardFrom is called to indicate a keepalive message (or equivalent) is received from a server
+func (hb *HeartBeatTracker) HeardFrom(id string, reset bool) {
+	hb.lock()
+	defer hb.unlock()
+
+	hb.last[id] = time.Now()
+}
+
+// TimedOut returns true if it is determined that have not heard from the server
+func (hb *HeartBeatTracker) TimedOut(id string) bool {
+	hb.lock()
+	defer hb.unlock()
+
+	t, ok := hb.last[id]
+	return !ok || time.Since(t) > hb.interval
+}
+
+// AverageTracker keeps track of the average latency of all messages.
+// Timeout: last received is more than the 'factor' of current average.
+type AverageTracker struct {
+	ch     chan struct{}
+	rec    map[string]averageTrackerRecord
+	factor int
+}
+
+type averageTrackerRecord struct {
+	cnt     int64
+	last    time.Time
+	totalMS int64 // in ms
+}
+
+func (rec *averageTrackerRecord) avg() int64 {
+	return rec.totalMS / rec.cnt
+}
+
+// NewAverageTracker returns a AverageTracker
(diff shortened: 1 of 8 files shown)


dfc/keepalivetracker.go
```diff
@@ -0,0 +1,129 @@
...
+// ideally this file stays under dfcpub, but currently dfc has a flat structure, following that
+// for now.
+import (
+	"time"
+)
+
+// HeartBeatTracker tracks timestamp of the last a message is received from a server.
+// Timeout: a message is not received within the interval.
+type HeartBeatTracker struct {
+	ch       chan struct{}
```
liangdrew: Any reason you prefer this over `sync.Mutex`?

dfc/keepalivetracker.go
```diff
@@ -0,0 +1,129 @@
...
+
+func (hb *HeartBeatTracker) lock() {
+	<-hb.ch
+}
+
+func (hb *HeartBeatTracker) unlock() {
+	hb.ch <- struct{}{}
+}
+
+// HeardFrom is called to indicate a keepalive message (or equivalvent) is received from a server
```
liangdrew: equivalvent -> equivalent

dfc/keepalivetracker.go
```diff
@@ -0,0 +1,129 @@
...
+	cnt     int64
+	last    time.Time
+	totalMS int64 // in ms
+}
+
+func (rec *averageTrackerRecord) avg() int64 {
+	return rec.totalMS / rec.cnt
+}
+
+// NewAverageTracker returns a AcerageTracker
```
liangdrew: AcerageTracker -> AverageTracker

dfc/keepalivetracker.go
```diff
@@ -0,0 +1,129 @@
+/*
```
liangdrew: Is this header necessary?

dfc/keepalivetracker.go
```diff
@@ -0,0 +1,129 @@
...
+	t, ok := hb.last[id]
+	return !ok || time.Since(t) > hb.interval
+}
+
+// AverageTracker keeps track of the average latency of all messages.
+// Timeout: last received is more than the 'factor' of current average.
+type AverageTracker struct {
+	ch     chan struct{}
+	rec    map[string]averageTrackerRecord
+	factor int
```
liangdrew: `factor` must be a whole number?

dfc/keepalivetracker.go
```diff
@@ -0,0 +1,129 @@
...
+
+	if rec.cnt == 0 {
+		return false
+	}
+
+	return int64(time.Now().Sub(rec.last)/time.Millisecond) > int64(a.factor)*rec.avg()
+}
+
+var (
+	_ KeepaliveTracker = &HeartBeatTracker{}
```
liangdrew: Purpose of these?

dfc/keepalivetracker.go
usmanong: avoid lock as much as possible, that's what i believe

dfc/keepalivetracker.go
usmanong: thanks, done

dfc/keepalivetracker.go
usmanong: done, thanks

dfc/keepalivetracker.go
usmanong: to make sure HeartBeatTracker and AverageTracker implements the KeepaliveTracker interface, compile time check

dfc/keepalivetracker.go
usmanong: not sure, can be changed later if necessary

dfc/keepalivetracker.go
usmanong: you mean the copy right?
every files has one.
it is not enough, i think later if doc is open sourced, we have to add more text approved by nvidia's legal to describe the license terms

dfc/keepalivetracker.go
VladimirMarkelov: Can rec.cnt be null?

usmanong: yes, but avg() is only called after it checked rec.cnt == 0
that's always the debate of how much to check.

PR #39, opened 2018-05-22
Track call stats between proxies and targets

usmanong:
thank you

### DIFF
--- a/dfc/callstats.go
+++ b/dfc/callstats.go
@@ -1,0 +1,109 @@
+/*
+ * Copyright (c) 2018, NVIDIA CORPORATION. All rights reserved.
+ */
+
+package dfc
+
+// Keep track average latency of call between proxies and targets
+// Currently it filters the calls by request (configurable), other filters (for example,
+// by http method) can be added if need to
+// A warning is logged if a call's latency exceeds the average by a predefined factor/threshhold
+// APIs provided by this package are concurrent access safe
+
+import (
+	"strings"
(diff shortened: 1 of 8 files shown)


dfc/callstats.go
```diff
@@ -0,0 +1,107 @@
...
+		return 0
+	}
+
+	return int64(l.total) / l.cnt
+}
+
+// NewCallStatsServer returns a CallStatsServer
+func NewCallStatsServer(requestsIncluded []string, factor float32) *CallStatsServer {
+	return &CallStatsServer{
+		ch:              make(chan callInfo, 100),
```
liangdrew: Why 100?

usmanong: just a number I picked, any suggestions?

liangdrew: Not sure what the optimal number here is - too low and it'll starting blocking. I would suggest to make it a const and give it a contextual comment so that it's easy to modify in the future if necessary.

usmanong: const is not necessary, comment I will add inside call stats.go, good?

liangdrew: Ok!

VladimirMarkelov: to me, worker function looks very fast, so I doubt 100 is not enough. E.g, keepalive happens once in a few tens of second

usmanong: I was thinking to make it in config file, but the config file is already huge

PR #40, opened 2018-05-22
Clean up tests

liangdrew:
1. Remove unused httpclient from getClusterMap()
2. Replace basic t.Fatal() with checkFatal()
3. Defer DestroyLocalBucket where not done previously
4. Use better corrupted xattr name for test

### DIFF
--- a/dfc/tests/regression_test.go
+++ b/dfc/tests/regression_test.go
@@ -294,7 +294,7 @@ func TestGetCorruptFileAfterPut(t *testing.T) {
 		}
 		return nil
 	})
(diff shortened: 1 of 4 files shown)


dfc/tests/regression_test.go
```diff
@@ -1439,12 +1419,9 @@ func getDaemonStats(httpclient *http.Client, t *testing.T, URL string) (stats ma
 	return
 }
 
-func getClusterMap(httpclient *http.Client, t *testing.T) dfc.Smap {
```
VladimirMarkelov: nice catch

VladimirMarkelov:
Why did you replace seeds?
As far as I remember, constant seeds were chosen to easily reproduce the same flow if something specific fails. Yes, it tests less values - only fixed set of them - but it generates a stable result.

PR #41, opened 2018-05-22
Fix file corruption used in test

liangdrew:


dfc/tests/regression_test.go
```diff
@@ -255,6 +255,9 @@ func TestGetCorruptFileAfterPut(t *testing.T) {
 		fqn        string
 	)
 	bucket := TestLocalBucketName
+	if bucket == "cloud" { // The corruption mechanism used in this test depends on this being false
```
usmanong: is this check needed?

dfc/tests/regression_test.go
```diff
@@ -274,6 +277,9 @@ func TestGetCorruptFileAfterPut(t *testing.T) {
 	// Test corrupting the file contents
 	f := <-filenameCh
 	filepath.Walk(rootDir, func(path string, info os.FileInfo, err error) error {
+		if info.IsDir() && info.Name() == "cloud" {
```
usmanong: don't see how is this related to the failure
the test uses file read from the channel, even this "cloud" happens, it still needs to match Base(path).
may be add an explanation in the pr to let reviews know what is the problem, and what is the fix

usmanong: never mind about this comment, I got it, still uncleaned file issue but not related to your tests. the cloud bucket has uncleaned files.
this pr is good to me, feel free to merge after addressing the first comment about check bucket name.

PR #42, opened 2018-05-22
Writes out call stats and keepalive stats to statsd

usmanong:

### DIFF
--- a/dfc/callstats.go
+++ b/dfc/callstats.go
@@ -15,6 +15,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/NVIDIA/dfcpub/dfc/statsd"
+
 	"github.com/golang/glog"
 )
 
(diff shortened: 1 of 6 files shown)


dfc/callstats.go
```diff
@@ -92,16 +97,49 @@ func (c *CallStatsServer) worker() {
...
 			if float64(ci.latency) > float64(c.factor)*float64(s.avg()) {
 				glog.Warningf("call %v latency %v too high, avg = %v", ci.url, ci.latency, s.avg())
 			}
 		} else {
-			c.stats[ci.url] = &latency{1, ci.latency}
+			c.stats[ci.url] = &latency{1, ci.latency, errCnt}
+		}
+
+		// from http://10.112.76.11:8080/v1/cluster/keepalive to 10_112_76_11:8080/v1/cluster/keepalive
+		// ideally 'url' is passed in as url.URL, which is not the case right now because
```
liangdrew: You can also use https://golang.org/pkg/net/url/#ParseRequestURI if you want a `url.URL`

usmanong: understand, my point was the whole system should have done using url.URL when possible, not translate back and forth.

PR #43, opened 2018-05-22
Verify daemon ID on target side before PUT

liangdrew:
This errors out object PUT requests to a target which do not provide a valid daemon ID (one that belongs to the cluster) as a query string parameter. Also add a test.


dfc/tests/target_test.go
```diff
@@ -0,0 +1,35 @@
...
+	smap, err := client.GetClusterMap(proxyurl)
+	checkFatal(err, t)
+
+	for sid = range smap.Tmap {
+		break
+	}
+
+	url := smap.Tmap[sid].DirectURL + "/" + dfc.Rversion + "/" + dfc.Robjects + "/" + bucket + "/" + object
+	err = client.HTTPRequest(http.MethodPut, url, nil)
+	if err == nil {
```
liangdrew: Ideally, `client.HTTPRequest` would return some response object instead of just an `error` so that more things (ex. status code) can be checked. Without doing string parsing (which is not a reliable way to test things), I think this is more or less the best we can get.

VladimirMarkelov: The similar thing I still have in backlog - make calls to DFC from client return normal HTTP error code. Otherwise it is hard to tell normal error, like object not found, from vague error, like internal server error

dfc/target.go
```diff
@@ -486,6 +486,22 @@ func (t *targetrunner) httpobjput(w http.ResponseWriter, r *http.Request) {
...
 	} else {
 		// PUT
+		d := query.Get(URLParamDaemonID)
+		b := false
+		for id := range t.smap.Pmap {
+			if id == d {
+				b = true
+			}
+		}
+		for id := range t.smap.Tmap {
```
VladimirMarkelov: When does target id appear in url? I see only proxy adds its id, so tmap lookup can be skipped.

liangdrew: Yup, forgot to remove this part. It was there originally for `sendfile()` which goes from target-to-target. Thanks!

dfc/proxy.go
```diff
@@ -607,7 +607,8 @@ func (p *proxyrunner) httpobjput(w http.ResponseWriter, r *http.Request) {
 		p.invalmsghdlr(w, r, errstr)
 		return
 	}
-	redirecturl := fmt.Sprintf("%s%s?%s=%t", si.DirectURL, r.URL.Path, URLParamLocal, p.islocalBucket(bucket))
+	redirecturl := fmt.Sprintf("%s%s?%s=%t&%s=%s", si.DirectURL, r.URL.Path, URLParamLocal,
+		p.islocalBucket(bucket), URLParamDaemonID, p.httprunner.si.DaemonID)
```
usmanong: what about other places that call target put? like rebalancing? I don't know the answer myself, just asking.

dfc/target.go
```diff
@@ -486,6 +486,17 @@ func (t *targetrunner) httpobjput(w http.ResponseWriter, r *http.Request) {
 		}
 	} else {
 		// PUT
+		d := query.Get(URLParamDaemonID)
+		b := false
+		for id := range t.smap.Pmap {
+			if id == d {
+				b = true
```
usmanong: break?

dfc/proxy.go
liangdrew: Good question - that case is handled here: https://github.com/NVIDIA/dfcpub/pull/43/files#diff-55511de5ecebac432bae78861a0686a5R477

liangdrew:
In the case of create local bucket, the target-side handler doesn't accept that action message (it defaults to the `invalmsghdlr` case. For situations similar to target PUT, we'd probably want to do something similar to that which is done here.

PR #46, opened 2018-05-23
Misc fixes including crashes on daemon start and stop due to empty po…

usmanong:
…inters

@VladimirMarkelov @sasanap @liangdrew @alex-aizman

### DIFF
--- a/dfc/callstats.go
+++ b/dfc/callstats.go
@@ -78,6 +78,10 @@ func (c *CallStatsServer) Stop() {
 }
 
 // Call sends a call's info to the channel
+// Note: L. Ding.
+// when httprunner is stopped, it doesn't mean there are no more intra-cluster calls anymore, those
+// calls will come to here and can trigger a write to closed channel error.
+// Unlikely but can happen.
 func (c *CallStatsServer) Call(url string, latency time.Duration, failed bool) {
 	c.ch <- callInfo{url, latency, failed}
 }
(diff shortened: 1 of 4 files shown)


dfc/callstats.go
```diff
@@ -73,13 +74,16 @@ func (c *CallStatsServer) Start() {
 
 // Stop stops the worker
 func (c *CallStatsServer) Stop() {
+	c.closed = true
```
liangdrew: Do you need to modify this atomically?

usmanong: good point, may be, let me check, thanks

usmanong: I probably do, but I would hate to do just for that, I will leave it as is, hope you are ok with it.
if it crashes on close, be it.

liangdrew: It should be relatively painless with a compare-and-swap operation on a uint, but up to you!

usmanong: the problem this is addressing should even be here, the root of the problem is dfc runners stop has no sequence, one example is after httprunner is stopped, there can still be calls pending, when they finish, they will call call stats, if the channel is closed too early, it causes an error, that's why I added this, in typical go code this should not happen.

usmanong: atomic is costly

usmanong: I just remove the closed flag, for now, let's see how often it happens,
my solution is half baked, probably as 'good' as no solution.

PR #48, opened 2018-05-24
remove unused field

VladimirMarkelov:
Primary field in proxyInfo was not used and was a source of confusion.
Removing the field made proxyInfo struct the same as struct daemonInfo, so the next step was to replace proxyInfo with daemonInfo.

Proxy tests passed.


usmanong:
very nice clean up
merge it

PR #50, opened 2018-05-24
Fix various races

usmanong:
1. Race during target runner start and stats runner log
2. Race during proxy start and keepalive runner
3. Race during target/proxy start and keepalive's first register call

### DIFF
--- a/dfc/keeper.go
+++ b/dfc/keeper.go
@@ -208,13 +208,19 @@ func (r *proxykalive) keepalive(err error) (stopped bool) {
 		glog.Infof("proxy %s keepalive triggered by err %v", r.p.si.DaemonID, err)
 	}
 
+	smapLock.Lock()
+
 	if r.p.primary {
+		smapLock.Unlock()
 		return r.primarykeepalive()
 	}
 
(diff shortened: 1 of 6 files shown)


dfc/keeper.go
```diff
@@ -208,11 +208,15 @@ func (r *proxykalive) keepalive(err error) (stopped bool) {
...
+
 	if r.p.primary {
+		smapLock.Unlock()
 		return r.primarykeepalive()
 	}
 
 	if r.p.smap.ProxySI == nil || !r.timedOut(r.p.smap.ProxySI.DaemonID) {
+		smapLock.Unlock()
 		return
 	}
```
VladimirMarkelov: Why smapLock is not Unlocked here?

usmanong:
actually hod on to this review, let me run all tests, sorry forgot to run it

dfc/keeper.go
usmanong: yup, that's why i think i need to run the whole tests, i only verified that all daemon starts
great eyes, that's why my code need reviews, thanks

usmanong:
ok, please review, cleared all tests
thanks

PR #52, opened 2018-05-25
Enable clients to use dfc as an HTTP proxy

liangdrew:
Clients can `export http_proxy=<proxy-url>`and this will let them start using dfc immediately.

Manual testing:

1. Download https://github.com/NVlabs/dlinputs.
2. `$ export http_proxy=http://localhost:8080`
2. `$ ./show-input http://storage.googleapis.com/lpr-ocr/uw3-lines.tgz`

### DIFF
--- a/dfc/gcp.go
+++ b/dfc/gcp.go
@@ -24,6 +24,8 @@ import (
 )
 
 const (
+	gcsURL = "http://storage.googleapis.com"
+
 	gcpDfcHashType = "x-goog-meta-dfc-hash-type"
 	gcpDfcHashVal  = "x-goog-meta-dfc-hash-val"
 
--- a/dfc/proxy.go
+++ b/dfc/proxy.go
@@ -190,7 +190,12 @@ func (p *proxyrunner) run() error {
 	p.httprunner.registerhdlr("/"+Rversion+"/"+Rhealth, p.httphealth)
 	p.httprunner.registerhdlr("/"+Rversion+"/"+Rvote+"/", p.votehdlr)
 	p.httprunner.registerhdlr("/"+Rversion+"/"+Rtokens, p.tokenhdlr)
-	p.httprunner.registerhdlr("/", invalhdlr)
+
+	if ctx.config.Net.HTTP.UseAsProxy {
+		p.httprunner.registerhdlr("/", p.reverseProxyHandler)
+	} else {
+		p.httprunner.registerhdlr("/", invalhdlr)
+	}
 
 	glog.Infof("Proxy %s is ready, primary=%t", p.si.DaemonID, p.primary)
 	glog.Flush()
@@ -485,7 +490,16 @@ func (p *proxyrunner) httpobjget(w http.ResponseWriter, r *http.Request) {
 		glog.Infof("passthru=false: proxy initiates the GET %s/%s", bucket, objname)
 		p.receiveDrop(w, r, redirecturl) // ignore error, proceed to http redirect
 	}
-	http.Redirect(w, r, redirecturl, http.StatusMovedPermanently)
+	if ctx.config.Net.HTTP.UseAsProxy {
+		req, err := http.NewRequest(http.MethodGet, redirecturl, r.Body)
+		if err != nil {
+			p.invalmsghdlr(w, r, err.Error())
+			return
+		}
+		p.revProxy.ServeHTTP(w, req)
+	} else {
+		http.Redirect(w, r, redirecturl, http.StatusMovedPermanently)
+	}
 
 	// Note: ideally, would prefer to call statsd in statsif.add(), but it doesn't have type support,
 	//       the name schema is different(statd doesn't need dup 'numget', it is get.count), statsif's dependency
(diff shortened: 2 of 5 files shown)


dfc/proxy.go
```diff
@@ -1613,6 +1626,32 @@ func (p *proxyrunner) tokenhdlr(w http.ResponseWriter, r *http.Request) {
...
+// handler for dfc being used as a HTTP proxy
+func (p *proxyrunner) proxyhdlr(w http.ResponseWriter, r *http.Request) {
+	baseURL := r.URL.Scheme + "://" + r.URL.Host
+	switch baseURL {
+	case gcpApiUrl:
+		switch r.Method {
+		case http.MethodGet:
+			s := p.restAPIItems(r.URL.Path, -1)
+			if len(s) == 2 {
+				r.URL.Path = "/" + Rversion + "/" + Robjects + r.URL.Path
```
VladimirMarkelov: There is URLPath function in httpcommon.go to make URL from the list

dfc/gcp.go
```diff
@@ -24,6 +24,8 @@ import (
 )
 
 const (
+	gcpApiUrl = "http://storage.googleapis.com"
```
usmanong: gcpAPIURL I think that's the go way, but not sure it looks cool in this case, what about urlGoogleAPI?

dfc/setup/config.sh
```diff
@@ -109,7 +109,8 @@ $FSPATHS
 	"callstats": {
 		"request_included": [ "keepalive", "metasync" ],
 		"factor": 2.5
-	}
+	},
+	"using_http_proxy": false
```
VladimirMarkelov: Maybe put the option into a group? 'netconfig'->'http'->'use_proxy'?

dfc/proxy.go
```diff
@@ -1613,6 +1626,32 @@ func (p *proxyrunner) tokenhdlr(w http.ResponseWriter, r *http.Request) {
...
+				p.httpobjget(w, r)
+			} else if len(s) == 1 {
+				r.URL.Path = "/" + Rversion + "/" + Rbuckets + r.URL.Path
+				p.httpbckget(w, r)
+			} else {
```
usmanong: use 'fallthrough' here and in default?

dfc/gcp.go
liangdrew: how about gcsURL? gcs = Google Cloud Storage (after all, this URL is specifically for GCS)

dfc/proxy.go
liangdrew: Can't `fallthrough` in the `else` or in a `default`, but I can `return` early in the first two cases and `fallthrough` otherwise

usmanong: us eat if possible, I was thinking return and fallthrough, see if the code looks good and decide

dfc/gcp.go
usmanong: sure

dfc/proxy.go
usmanong: may be don't use switch for now at all because there is only one case

dfc/proxy.go
```diff
@@ -1613,6 +1626,33 @@ func (p *proxyrunner) tokenhdlr(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+// handler for dfc being used as a HTTP proxy
+func (p *proxyrunner) proxyhdlr(w http.ResponseWriter, r *http.Request) {
```
usmanong: this name will be very confusing for people read the code for the first time, including both "reverse proxy" probably better

dfc/proxy.go
```diff
@@ -1636,8 +1676,8 @@ func (p *proxyrunner) httpclugetstats(w http.ResponseWriter, r *http.Request, ge
 	for _, si := range p.smap.Tmap {
 		stats := &storstatsrunner{Capacity: make(map[string]*fscapacity)}
 		out.Target[si.DaemonID] = stats
-		url := si.DirectURL + "/" + Rversion + "/" + Rdaemon
-		res := p.call(r, si, url, r.Method, getstatsmsg)
+		u := si.DirectURL + URLPath(Rversion, Rdaemon)
```
usmanong: thank you

dfc/proxy.go
```diff
@@ -485,7 +490,15 @@ func (p *proxyrunner) httpobjget(w http.ResponseWriter, r *http.Request) {
 		glog.Infof("passthru=false: proxy initiates the GET %s/%s", bucket, objname)
 		p.receiveDrop(w, r, redirecturl) // ignore error, proceed to http redirect
 	}
-	http.Redirect(w, r, redirecturl, http.StatusMovedPermanently)
+	if ctx.config.Net.HTTP.UseAsProxy {
+		req, err := http.NewRequest(http.MethodGet, redirecturl, r.Body)
+		if err != nil {
+			p.invalmsghdlr(w, r, err.Error())
```
VladimirMarkelov: return after error?

liangdrew: Yes, thanks, added

PR #53, opened 2018-05-25
Fix multiple response.WriteHeader calls bug

liangdrew:
io.CopyBuffer() implicitly calls WriteHeader()

### DIFF
--- a/dfc/httpcommon.go
+++ b/dfc/httpcommon.go
@@ -667,8 +667,13 @@ func (h *httprunner) setconfig(name, value string) (errstr string) {
 //=================
 func (h *httprunner) invalmsghdlr(w http.ResponseWriter, r *http.Request, specific string, other ...interface{}) {
 	status := http.StatusBadRequest
+	shouldWriteError := true
 	if len(other) > 0 {
-		status = other[0].(int)
+		if v, ok := other[0].(int); ok {
+			status = v
+		} else if v, ok := other[0].(bool); ok {
+			shouldWriteError = v
+		}
 	}
 	s := http.StatusText(status) + ": " + specific + ": " + r.Method + " " + r.URL.Path
 	if _, file, line, ok := runtime.Caller(1); ok {
@@ -679,7 +684,9 @@ func (h *httprunner) invalmsghdlr(w http.ResponseWriter, r *http.Request, specif
 	}
 	glog.Errorln(s)
 	glog.Flush()
-	http.Error(w, s, status)
+	if shouldWriteError {
+		http.Error(w, s, status)
+	}
 	h.statsif.add("numerr", 1)
 }
 
--- a/dfc/target.go
+++ b/dfc/target.go
@@ -426,7 +426,7 @@ existslocally:
 	written, err := io.CopyBuffer(w, file, buf)
 	if err != nil {
 		errstr = fmt.Sprintf("Failed to send file %s, err: %v", fqn, err)
-		t.invalmsghdlr(w, r, errstr)
+		t.invalmsghdlr(w, r, errstr, false)
 		return
 	}
 	if !coldget {


dfc/target.go
```diff
@@ -426,7 +426,7 @@ existslocally:
 	written, err := io.CopyBuffer(w, file, buf)
 	if err != nil {
 		errstr = fmt.Sprintf("Failed to send file %s, err: %v", fqn, err)
-		t.invalmsghdlr(w, r, errstr)
+		t.invalmsghdlr(w, r, errstr, false)
```
VladimirMarkelov: Nice catch but it looks kind of hacky :) And if one wants skip writing to header and set error code it does not help. Maybe just replace this line with two:
```
glog.Errorln(errstr)
t.statsif.add("numerr", 1)
```
? In case of 'false' almost all of the invalmsghdlr does nothing

dfc/httpcommon.go
```diff
@@ -667,8 +667,13 @@ func (h *httprunner) setconfig(name, value string) (errstr string) {
 //=================
 func (h *httprunner) invalmsghdlr(w http.ResponseWriter, r *http.Request, specific string, other ...interface{}) {
 	status := http.StatusBadRequest
+	shouldWriteError := true
 	if len(other) > 0 {
-		status = other[0].(int)
+		if v, ok := other[0].(int); ok {
```
usmanong: this probably not a good idea, the API changed from (w, r, errstr, status) to (w, r, errstr, bool).
have you checked all caller of invalmsghdlr()? are they all called with one or none parameters? if that's the case, which I think it is, based on the fact that it only checked other[0], then I would make this more explicit, as (w, r, errstr, status, writeErr), seems a lot change though.
or keep it this way and document it clearly.
the existing API only try to avoid passing status, explicit calls seems better, a wrapper like can take care of them
and the name invalmsghdlr is really bad to me.
but any way, you can keep it your way, from a bad API to slight worse API, fine.

dfc/target.go
usmanong: so this means all places that io.CopyBuffer() is called and failed, no need to write out a http error because it is already failed?
nice finding

VladimirMarkelov: 1. Invalmsghdlr traces only caller. So, moving glog to here is the same as keep calling invalmsghdlr with false - in both cases glog ouputs target.go:429. As for URL - you can print it in-place :)
2. I do not think changing a function to fit only one corner-case call is better than replacing the call with 2 lines of code for that corner case

VladimirMarkelov:
Yes, I was thinking about adding a new function that only reports about error but does not write anything to headers. It won't be a big change - the current invalmsghdlr stays the same, place with CopyBuffer will call new "invalimsghdlr". As Python Zen says 'Explicit is better than implicit.' :)

usmanong:
new function is fine, the old API should have been there in the first place.
see how much code you have to duplicate

VladimirMarkelov:
If it was not stack trace with 'Caller', the old function could have called new one and then wrote to headers

dfc/target.go
liangdrew: The issue is actually that a new status code _can't_ be written to the writer because that header has already been written by `io.CopyBuffer` (since it calls `Write()`, which calls `WriteHeader()`). If `WriteHeader()` is called again, it returns a long and messy error.

liangdrew:
I can add a new function just to log the error, but this is not very clean either since the error must be returned by this new function so that `http.Error()` can use it as an argument in `invalmsghdlr()`. I'll look at slightly bigger refactors to unwind a bit of this mess.

usmanong:
what about a new function A() which has the code currently in invalmsghdlr() like:
A(w, r, errstr, status, writeErr) {
	s := http.StatusText(status) + ": " + specific + ": " + r.Method + " " + r.URL.Path
	if _, file, line, ok := runtime.Caller(1); ok {
		if !strings.Contains(specific, ".go, #") {
			f := filepath.Base(file)
			s += fmt.Sprintf("(%s, #%d)", f, line)
		}
	}
	glog.Errorln(s)
	glog.Flush()
	http.Error(w, s, status)
	h.statsif.add("numerr", 1)
}

the your new function calls like:
A(w, r, errstr, http.badrequest, false)

and invalmsghdlr calls like:
A(w, r, errstr, status, true)

liangdrew:
I don't like having to add those two extra arguments to every call to this function, but you're right, that works.

usmanong:
no one should call A() directly, they call your new function and the existing invalmsghdlr(), so nothing changes outside

liangdrew:
Oh I see what you mean - having two wrappers for A(). I like that, but I also feel bad about two functions which only call other functions. If I can't think of anything better, I'll probably do this.

PR #54, opened 2018-05-27
Primary proxy start up logic change and unit tests for it

usmanong:


dfc/proxy.go
```diff
@@ -325,50 +331,38 @@ func (p *proxyrunner) findRole() {
...
+		return
+	}
 
-		if time.Now().After(deadline) {
-			glog.Warning("findRole took too long")
-			return
-		}
+	smapLock.Lock()
+	if smap.ProxySI.DaemonID == p.si.DaemonID {
+		glog.Infoln("This proxy is primary in found cluster map; remaining primary.")
```
VladimirMarkelov: I did not check: is it possible that discover returns smap with this proxy as primary but at this moment p.primary is false?

usmanong: not possible, this path is for a proxy to start as primary and trying to figuring out what is the latest map in order to decide whether it should give up the rights to be primary and down grade itself to non primary

PR #56, opened 2018-05-29
Fix multiple response.WriteHeader calls bug

liangdrew:
[io.CopyBuffer()](https://github.com/NVIDIA/dfcpub/compare/zero-change-debug?expand=1#diff-55511de5ecebac432bae78861a0686a5R426) implicitly calls `WriteHeader()`. [http.Error()](https://github.com/NVIDIA/dfcpub/compare/zero-change-debug?expand=1#diff-1ae6ff57d44b5dff7d5b9b647c4a75b9R687) calls `WriteHeader()` as well, so if an error was returned from `io.CopyBuffer()`, a `multiple response.WriterHeader calls` error occurred.

### DIFF
--- a/dfc/target.go
+++ b/dfc/target.go
@@ -426,7 +426,8 @@ existslocally:
 	written, err := io.CopyBuffer(w, file, buf)
 	if err != nil {
 		errstr = fmt.Sprintf("Failed to send file %s, err: %v", fqn, err)
-		t.invalmsghdlr(w, r, errstr)
+		glog.Errorln(t.richHTTPError(r, errstr, http.StatusInternalServerError, 1))
+		t.statsif.add("numerr", 1)
 		return
 	}
 	if !coldget {
--- a/dfc/httpcommon.go
+++ b/dfc/httpcommon.go
@@ -665,18 +665,25 @@ func (h *httprunner) setconfig(name, value string) (errstr string) {
 // http err + spec message + code + stats
 //
 //=================
-func (h *httprunner) invalmsghdlr(w http.ResponseWriter, r *http.Request, specific string, other ...interface{}) {
-	status := http.StatusBadRequest
-	if len(other) > 0 {
-		status = other[0].(int)
-	}
-	s := http.StatusText(status) + ": " + specific + ": " + r.Method + " " + r.URL.Path
-	if _, file, line, ok := runtime.Caller(1); ok {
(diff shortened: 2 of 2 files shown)


dfc/target.go
```diff
@@ -426,7 +426,8 @@ existslocally:
 	written, err := io.CopyBuffer(w, file, buf)
 	if err != nil {
 		errstr = fmt.Sprintf("Failed to send file %s, err: %v", fqn, err)
-		t.invalmsghdlr(w, r, errstr)
+		glog.Errorln(t.richHTTPError(r, errstr, http.StatusInternalServerError))
+		t.statsif.add("numerr", 1)
```
usmanong: this means each call to richHTTPError() has to call this?
as i mentioned last week, what i was thinking is do something like this:
fun nameForNewErrorFunc(w, r, errstr, status, ) {
   does everything existing invalmsghdlr() does but without the part that collect stats
	if len(other) > 0 {
 		status = other[0].(int)
 	}
}

so existing invalmsghdlr() becomes:
{ 
   extract status and call nameForNewErrorFunc()
}

and your new fund simply call nameForNewErrorFunc() with the right status and error

dfc/target.go
```diff
@@ -426,7 +426,8 @@ existslocally:
 	written, err := io.CopyBuffer(w, file, buf)
 	if err != nil {
 		errstr = fmt.Sprintf("Failed to send file %s, err: %v", fqn, err)
-		t.invalmsghdlr(w, r, errstr)
+		glog.Errorln(t.richHTTPError(r, errstr, http.StatusInternalServerError, 1))
```
VladimirMarkelov: Just an idea to discuss: what if we remove skip from arguments and use a constant skip=2?
1. No need to pass 'skip' - except this line 2 is used
2. As for this line, even with skip=2 we have a) target.go:429 in log because glog ouputs its line and b) at the same time, we will know who called 'existslocally' because of skip=2 show it - that can be sometimes more useful than to know that utility function 'existslocally' failed

liangdrew: It would work fine for now, but it may have to be increased at some point in the future. The thing I don't like is that it is not explicit - users of the function would have to remember that the function _only_ skips two stack frames.

liangdrew: Also, the file and line # appended to the end of the error string will always be the caller from two stack frames up. This could be misleading at times.

PR #116, opened 2023-01-18
Invalidate cached object based on version/chksum

### LINKED ISSUE #114: Ability to get last version from backend storage
I want use ais as cache tier with another ais cluster as provider. How can i validate that object in cache is not obsolete? What about extend get object api and add object version as query param. If that version not found in cache then initiate cold get from backend?

and-1:
Get object from remote cluster if requested objects version/chksum  does not exist in cache and found in remote

Resolve #114

### DIFF
--- a/api/apc/query.go
+++ b/api/apc/query.go
@@ -64,6 +64,12 @@ const (
 	// force the operation; allows to overcome certain restrictions (e.g., shutdown primary and the entire cluster)
 	// or errors (e.g., attach invalid mountpath)
 	QparamForce = "frc"
+
+	//Set version in query param to invalidate warm cache if not match working with remote ais cluster
+	QparamWarmVersion = "warm_version"
+
+	//Set checksum in query param to invalidate warm cache if not match working with remote ais cluster
+	QparamWarmChkSum = "warm_chksum"
 )
 
 // QparamFltPresence enum.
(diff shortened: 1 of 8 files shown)


api/apc/query.go
```diff
@@ -64,6 +64,12 @@ const (
 	// force the operation; allows to overcome certain restrictions (e.g., shutdown primary and the entire cluster)
 	// or errors (e.g., attach invalid mountpath)
 	QparamForce = "frc"
+
+	//Set version in query param to invalidate warm cache if not match working with remote ais cluster
+	QparamWarmVersion = "warm_version"
+
+	//Set checksum in query param to invalidate warm cache if not match working with remote ais cluster
+	QparamWarmChkSum = "warm_chksum"
```
alex-aizman, line 78: Not sure "warm" is the right naming. Maybe, "user"?

api/object.go
```diff
@@ -273,6 +273,44 @@ func GetObject(bp BaseParams, bck cmn.Bck, object string, options ...GetObjectIn
 	return resp.n, nil
 }
 
+// GetObjectReader returns reader of the requested object. It does not read body
+// bytes, nor validates a checksum. Caller is responsible for closing the reader.
+func GetObjectReaderWithOA(bp BaseParams, bck cmn.Bck, object string, options ...GetObjectInput) (r io.ReadCloser, oa *cmn.ObjAttrs, err error) {
+	var (
```
alex-aizman, line 279: I'll stop at this point. Take a look at CONTRIBUTING.md for general guidelines.

Again, lots of respect for the effort!

gaikwadabhishek:
This MR is being closed because there has been no activity for over 90 days.

PR #179, opened 2024-07-17
python (botocore patch): initialize s3_redirect context to avoid KeyError in redirection logic

rkoo19:
Signed-off-by: Ryan Koo <rbk65@cornell.edu>


rkoo19:
Had outdated dependency.

PR #193, opened 2024-12-04
ci: disable testifylint since pkg is not used

ferhatelmas:


alex-aizman:
What's the motivation behind removing this particular linter? Ok, it is not now but maybe it'll be used in the future.

It's a no-op, why should we remove it

ferhatelmas:
Thanks for checking.

> What's the motivation behind removing this particular linter?

Not needed, why to run something unnecessarily? It could save CPU cycles and indirectly prevents one dependency. 

> it is not used now but maybe it'll be used in the future

I would think adding at the same diff if it starts to be used. 

> It's a no-op, why should we remove it

A suggestion, feel free to ignore.

PR #196, opened 2024-12-30
workflows/test-short - Reorder steps to cache dependencies

ScottBrenner:
Reordering the steps here to address a warning on runs:

> Warning: Restore cache failed: Dependencies file is not found in /home/runner/work/aistore/aistore. Supported file pattern: go.sum

as the code is not yet checked out, reference https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows


alex-aizman:
LTGM

PR #197, opened 2025-01-01
GitHub Actions workflows - Reorder steps to cache dependencies

ScottBrenner:
Extending https://github.com/NVIDIA/aistore/commit/c48ef600c1c4f77e8e3fd8765397875bd5802c3c to other workflows in the repository

PR #198, opened 2025-01-01
Dependabot configuration to update actions in workflow

ScottBrenner:
Noticed a new version of actions/setup-go is available - https://github.com/actions/setup-go/tree/main?tab=readme-ov-file#v5 -  proposing a Dependabot configuration to update - reference https://docs.github.com/en/actions/security-guides/using-githubs-security-features-to-secure-your-use-of-github-actions#keeping-the-actions-in-your-workflows-secure-and-up-to-date

Suggest enabling https://docs.github.com/en/code-security/dependabot/working-with-dependabot/about-dependabot-on-github-actions-runners#enabling-or-disabling-for-your-repository as well

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+# and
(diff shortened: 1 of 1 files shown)


alex-aizman:
ditto https://github.com/NVIDIA/aistore/pull/197#issuecomment-2567071093

alex-aizman:
that's "dependabot version updates" looks like

PR #199, opened 2025-01-03
Extend Dependabot version updates to gomod package ecosystem

ScottBrenner:
Extending the Dependabot version updates here to the gomod package ecosystem, reference https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#package-ecosystem-

Example output on my fork -https://github.com/ScottBrenner/aistore/pulls?q=is%3Apr+is%3Aopen+label%3Ago

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,3 +9,7 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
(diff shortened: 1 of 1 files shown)


alex-aizman:
No, I'd rather not have any automated version updates - `go get` at the right times is good enough, and works.

Closing.

PR #216, opened 2025-04-11
Add alive_progress to dependencies

ghtyrant:
alive_progress is imported in aistore/pytorch/iter_dataset.py - without this dependency the examples fail to run

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -72,6 +72,7 @@ pytorch = [
     "torch",
     "torchdata",
     "webdataset>=0.2.100",
+    "alive_progress>=3.1.0"
 ]
 botocore = [
     "wrapt",


aaronnw:
Thanks for opening! 

The change looks good, can you take a look at our [CONTRIBUTING.md](https://github.com/NVIDIA/aistore/blob/main/CONTRIBUTING.md#signing-off-commits) and add a signature to the commit (git commit -s)?

python/pyproject.toml
```diff
@@ -53,6 +53,7 @@ dependencies = [
     "msgspec>=0.18.5",
     "xxhash>=3.3.0",
     "tenacity>=9.0.0",
+    "alive_progress>=3.1.0"
```
aaronnw, line 56: Can you actually move this down to the `pytorch` option dependencies section below? A quick check shows we don't use it outside the pytorch module.

aaronnw (changes requested):

gaikwadabhishek:
Is there a spelling mistake in the commit message?

Can you rename it to 
```python: add `alive_progress` for PyTorch plugin support```

if you are adding desc in the commit message try to wrap it and make it less than 120 chars per line. I would say skip the desc for this commit

python/pyproject.toml
ghtyrant: Done!

PR #217, opened 2025-04-13
replace Walk with WalkDir

Gofastasf:
* Avoid lstat call on every file accessed.

### DIFF
--- a/ais/utils.go
+++ b/ais/utils.go
@@ -426,7 +426,7 @@ func cleanupConfigDir(name string, keepInitialConfig bool) {
 		cos.RemoveFile(daemon.cli.localConfigPath)
 	}
 	config := cmn.GCO.Get()
-	filepath.Walk(config.ConfigDir, func(path string, finfo os.FileInfo, _ error) error {
+	filepath.WalkDir(config.ConfigDir, func(path string, finfo os.DirEntry, _ error) error {
 		if strings.HasPrefix(finfo.Name(), ".ais.") {
 			if err := cos.RemoveFile(path); err != nil {
 				nlog.Errorf("%s: failed to cleanup %q, err: %v", name, path, err)


alex-aizman (approved):
LGTM. Separately, I wonder why there's still no linter for this... since Go 1.16

Gofastasf:
The go team has not deprecated the Walk yet - https://github.com/golang/go/issues/69604

So the static analyzer author turned down to add linter rule for flagging Walk https://github.com/dominikh/go-tools/issues/1014

alex-aizman:
there's stuff about showing (or not showing) virtual directories and `ais ls --nr` execution that may require attention

Gofastasf:
Just to clarify was this already present before switching to WalkDir ?

alex-aizman:
I don't know, frankly. It's on my list though:

```sh
$ ais ls s3://data --nr
NAME                                             SIZE            CACHED  
stuff_manifest.json                          2.19MiB         yes     
other_stuff_manifest_lower.json                    2.14MiB         yes     
testfile                                         13B             yes    
 
$ ais ls s3://data/ASR --limit 1
NAME                                             SIZE            CACHED  
ASR/blah.wav   593.94KiB       yes     
```

Meaning, `ASR` is right there but it doesn't show up non-recursively. Something's changed. Extra slash in `ais ls s3://data/ --nr` doesn't help.

Gofastasf:
Hey I don't think this is specifically related to switching Walk to WalkDir as the only difference is Walk calls Lstat calls on every file traversed and WalkDir avoids it. 

So the possiblities I see are : 

- ASR is a non empty directory in local or global config path ? os.Remove doesn't remove non empty directory https://pkg.go.dev/os#Remove and we have ignored the error handling there. 

- Permission error or other accesibility errors causing them to not remove it as we have ignored error inside the walk. 

- ASR is a symlink ?

Can you try it with this code - 

```
func cleanupConfigDir(name string, keepInitialConfig bool) {
    if !keepInitialConfig {
        if err := cos.RemoveFile(daemon.cli.globalConfigPath); err != nil {
            nlog.Errorf("%s: Failed to remove global config %q: %v", name, daemon.cli.globalConfigPath, err)
        }
        if err := cos.RemoveFile(daemon.cli.localConfigPath); err != nil {
            nlog.Errorf("%s: Failed to remove local config %q: %v", name, daemon.cli.localConfigPath, err)
        }
    }
    config := cmn.GCO.Get()
    nlog.Infoln(name, "Cleaning config dir:", config.ConfigDir)
    err := filepath.WalkDir(config.ConfigDir, func(path string, finfo fs.DirEntry, err error) error {
        if err != nil {
            nlog.Errorf("%s: WalkDir error at %q: %v", name, path, err)
            return nil
        }
        if finfo.IsDir() {
            nlog.Infoln("%s: Visiting directory %q", name, path)
        }
        if finfo.Type()&os.ModeSymlink != 0 {
            nlog.Infoln("%s: Found symlink %q", name, path)
        }
        if strings.HasPrefix(finfo.Name(), ".ais.") {
            nlog.Infoln("%s: Removing file %q", name, path)
            if err := cos.RemoveFile(path); err != nil {
                nlog.Errorf("%s: Failed to cleanup %q: %v", name, path, err)
            }
        } else {
            nlog.Infoln("%s: Skipping non-.ais. file %q", name, path)
        }
        return nil
    })
    if err != nil {
        nlog.Errorf("%s: WalkDir failed: %v", name, err)
    }
}
```

Gofastasf:
This is unlikely though. I think it could be a path normalization issue somewhere on handling non recursive function or some error in its higher level traversal. 

Also is it s3 specific issue ?

I will take a look at this as soon as I get access to my machine.

Gofastasf:
I think I found the bug. It was related to S3 object prefix which was not recognized as virtual directories and was being skipped during the listing.

alex-aizman:
possibly so. First of, I'd use a single target, single proxy to reduce the amount of stuff to look at. Further, there are several points of interest but, as always, makes sense to start from the easiest.

* https://github.com/NVIDIA/aistore/blob/main/ais/backend/aws.go#L305   - and establish the fact that `--nr` option from CLI propagates all the way

* https://github.com/NVIDIA/aistore/blob/main/ais/backend/aws.go#L359   - add debug log to print zero-size entries that we'd expect to be appended and returned

Next, if all of the above is fine, I'd maybe completely skip target logic and check both CLI and proxy, the latter somewhere around `concatLso` that combines list-objects returns from all targets

PR #221, opened 2025-06-02
refactor(benchmark): replace b.N loops with b.Loop()

Gofastasf:
* Eliminates the need for explicit calls to b.ResetTimer and b.StopTimer for setup and cleanup.
* Prevents dead code elimination within the loop.

For reference: https://go.dev/blog/testing-b-loop

### DIFF
--- a/ais/tgtobj_internal_test.go
+++ b/ais/tgtobj_internal_test.go
@@ -111,8 +111,7 @@ func BenchmarkObjPut(b *testing.B) {
 				b.Fatal(err)
 			}
 
-			b.ResetTimer()
-			for range b.N {
+			for b.Loop() {
 				b.StopTimer()
 				r, _ := readers.NewRand(bench.fileSize, cos.ChecksumNone)
 				poi := &putOI{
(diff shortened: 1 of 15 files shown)


Gofastasf:
Remaining places b.N is used: 
https://github.com/NVIDIA/aistore/blob/main/cmn/cos/bytepack_test.go 
https://github.com/NVIDIA/aistore/blob/main/cmn/prob/dyn_cuckoo_test.go

Using b.N  for input sizing might be bad ([ref1](https://github.com/golang/go/issues/61515), [ref2](https://eli.thegreenplace.net/2023/common-pitfalls-in-go-benchmarking/)), this can cause: 
- Quadratic runtime growth (both iterations and input complexity scale with b.N)
- Inconsistent measurements across different b.N values
- Expensive setup repeated during ramp-up

Should I change these to fixed input size ? Or do you prefer to keep it this way ? Let me know what input sizes would make sense if we go with hardcoded values.

Gofastasf:
After switching to b.Loop, some benchmarks got significantly faster and more stable. For example b.Loop got ~4x faster for BenchmarkSend and ~20x faster for BenchmarkSendParallel.

alex-aizman:
LGTM

> Should I change these to fixed input size ?

Yes, with a comment.

PR #223, opened 2025-06-07
refactor: filter benchmarks; use NewDefaultFilter; update concurrent load bench

Gofastasf:
* use NewDefaultFilter to benchmark 
* update concurrent load benchmark for filter
* replaced b.N with b.Loop

### DIFF
--- a/cmn/prob/dyn_cuckoo_test.go
+++ b/cmn/prob/dyn_cuckoo_test.go
@@ -6,7 +6,9 @@ package prob_test
 
 import (
 	"math/rand/v2"
+	"runtime"
 	"sync"
+	"sync/atomic"
 	"testing"
 
 	"github.com/NVIDIA/aistore/cmn/prob"
@@ -17,8 +19,9 @@ import (
 )
 
 const (
-	testFilterInitSize = 100 * 1000
 	objNameLength      = 5
+	defaultNumKeys     = 10_000_000 // default filter is 10M in size
+	dynamicNumKeys     = 5 * 10_000_000
 )
 
 // Predefined buckets.
@@ -39,7 +42,7 @@ func genKeys(keysNum int) [][]byte {
 }
 
 var _ = Describe("Filter", func() {
-	filter := prob.NewFilter(testFilterInitSize)
+	filter := prob.NewDefaultFilter()
 
 	BeforeEach(func() {
 		filter.Reset()
@@ -53,7 +56,7 @@ var _ = Describe("Filter", func() {
 		})
 
 		It("should lookup the keys in filter with no more than 0.06% failure rate", func() {
-			keys := genKeys(testFilterInitSize * 10)
+			keys := genKeys(defaultNumKeys * 5)
 			total := float64(len(keys))
 			for _, key := range keys {
 				filter.Insert(key)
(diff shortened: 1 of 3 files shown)


cmn/prob/dyn_cuckoo_test.go
```diff
@@ -91,135 +93,137 @@ var _ = Describe("Filter", func() {
...
 		go func() {
-			for n := range b.N {
-				filter.Lookup(keys[n])
+			for i := range b.N {
+				filter.Lookup(keys[i%len(keys)])
 			}
 			wg.Done()
 		}()
 		go func() {
-			for n := range b.N {
```
Gofastasf, line 221: I'm not sure if we are getting an accurate benchmark here. Should we instead benchmark each of Insert, Lookup, Delete seperately under b.RunParallel ?

Gofastasf: Okay removed that benchmark.

alex-aizman:
instead of removing, do you want to compare vs regular binary mutex?

```diff
--- a/cmn/prob/dyn_cuckoo.go
+++ b/cmn/prob/dyn_cuckoo.go
@@ -28,7 +28,7 @@ const (
 type Filter struct {
        filters []*cuckoo.Filter
        size    uint
-       mtx     sync.RWMutex
+       mtx     sync.RWMutex // <<< add alternative Filter type that uses sync.Mutex, and compare BenchmarkInsertAndDeleteAndLookupParallel
 }
```

alex-aizman:
yup. So what do you suggest?

Keeping in mind that the only utilitarian purpose of this micro-bench us to help is make better decision wrt existing code in the `reb` package...

Gofastasf:
I think we can switch to sync.Mutex here as Lookup is followed by Delete anyways. And sync.RWMutex was reported to be 2x slower than sync.Mutex and scales poorly with CPU count 

https://github.com/golang/go/issues/38813
https://github.com/golang/go/issues/17973

alex-aizman:
2016 and 2020  - is old. But if you show some sort of adequate benchmark then - certainly.

Gofastasf:
Sure I'm looking into it.

Gofastasf:
This does backround insert and benchmark parallel lookups and conditional delete if the keys are send by the GFN. 

I have removed the old BenchmarkInsertAndDeleteAndLookupParallel benchmark. Let me know if you still need it.

Gofastasf:
A few more changes needed. I'm working on it.

Gofastasf:
Hey its finished. Now the insertion will work in backround.
But the same keys are being inserted on the filter. So there is chance for multiple lookup hits...

Gofastasf:
I think we need to benchmark default filter instead of small filter and preallocate with something higher than the default ?

alex-aizman:
yes, `prob.NewDefaultFilter` if that's what you mean. 

In fact, `prob.NewFilter` must be likely lowercased in hidden. The test must use  `prob.NewDefaultFilter` - the one that's used in the `reb` package.

Gofastasf:
Yes that's the one I meant. I will replace other benchmarks also to use default filter. So the current default filter size is 10M. Do we need to benchmark the dynamic growth with more than 10M keys?

alex-aizman:
LGTM

Incidentally, I ran `-bench=Concurrent` with a binary mutex:

```diff
+++ b/cmn/prob/dyn_cuckoo.go
@@ -28,7 +28,7 @@ const (
 type Filter struct {
        filters []*cuckoo.Filter
        size    uint
-       mtx     sync.RWMutex
+       mtx     sync.Mutex
 }
```

Got more than 2x degradation in perf.

PR #229, opened 2025-07-17
fix broken reference to benchmark

Gofastasf:


rkoo19:
Good catch, thank you @Gofastasf!

Gofastasf:
Seems like there are lot of other links broken too. Will try to fix it.

alex-aizman:
thank you!

PR #250, opened 2025-12-04
Fix status code for range reads

michelle-tan-ai:
### Issue
According to [RFC 9110](https://datatracker.ietf.org/doc/html/rfc9110#name-range), "If all of the preconditions are true, the server supports the Range header field for the target resource, the received Range field-value contains a valid ranges-specifier with a range-unit supported for that target resource, and that ranges-specifier is satisfiable with respect to the selected representation, the server SHOULD send a 206 (Partial Content) response with content containing one or more partial representations that correspond to the satisfiable range-spec(s) requested."

Issue repro:
`curl -L -H "Range: bytes=0-1023" -i <URL> --output data.txt`
Returns a proper range read response but without the 206 status code
```
HTTP/1.1 200 OK^M
Accept-Ranges: bytes^M
Content-Length: 1024^M
Content-Range: bytes 0-1023/33241197^M
```

This incorrect status code makes it not possible to open a [lance](https://lance.org/) dataset

### Fix
Returns a 206 status code

### DIFF
--- a/ais/test/object_test.go
+++ b/ais/test/object_test.go
@@ -1691,6 +1691,10 @@ func verifyValidRangesQuery(t *testing.T, proxyURL string, bck cmn.Bck, objName,
 		cos.HdrAcceptRanges, acceptRanges)
 	contentRange := respHeader.Get(cos.HdrContentRange)
 	tassert.Errorf(t, contentRange != "", "%q header should be set", cos.HdrContentRange)
+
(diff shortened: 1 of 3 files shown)


alex-aizman:
pushed:

https://github.com/NVIDIA/aistore/commit/718b2d9b050f2aafdb245526bffd437072102fb2

PR #261, opened 2026-02-21
fs: fix LPI test parse failure and wire randomized page size

ferhatelmas:
* trim wc -l whitespace before parsing
* use computed random page size to lpiPageSize as intended


alex-aizman:
LGTM - thanks!

PR #285, opened 2026-04-19
python/etl: stream FastAPI no-FQN hpull GET without full-body buffering

chanu1406:
FastAPI no-FQN `hpull` GET requests were buffering the full upstream object into memory before passing it to `transform_stream()`. That defeated streaming for large objects and diverged from the Flask and HTTP server implementations, which already stream chunk-by-chunk.

This PR narrows the fix to the FastAPI no-FQN `hpull` GET path:

- open the upstream GET with the shared sync `requests.Session`, wrapped in `asyncio.to_thread(...)` so the event loop stays responsive
- pass `transform_stream()` a sync file-like reader instead of a buffered `BytesIO`
- wrap the `requests.Response.raw` stream so `close()` goes through `resp.close()`, returning the connection to the pool on early-close and error paths
- forward upstream HTTP statuses as-is and keep genuine network failures mapped to 502

Out of scope:

- no-FQN PUT still buffers the request body today; this PR does not change PUT behavior

Tests:

- add unit coverage for the non-buffering FastAPI streaming GET path
- cover upstream status forwarding, network-error 502 handling, connection release on upstream error/close, and exact path forwarding
- expand unit coverage around `_direct_put_stream_with_retry` reader lifecycle and retry behavior

### DIFF
--- a/python/CHANGELOG.md
+++ b/python/CHANGELOG.md
@@ -15,6 +15,10 @@ We structure this changelog in accordance with [Keep a Changelog](https://keepac
   presence check, `max_resume` bound, restart-from-zero on eviction).
   Detection is skipped when the response carries a non-identity
   `Content-Encoding`.
+- ETL `FastAPIServer`: stream no-FQN `hpull` GETs in constant memory via the
+  shared `requests.Session`, wrapped in `asyncio.to_thread` to keep the event
+  loop responsive. Previously the full upstream object was buffered into memory
+  before being handed to `transform_stream`.
 - **ETL direct-put retry**: added exponential-backoff retry for transient connection
   errors in Flask and HTTP multi-threaded ETL servers for parity with FastAPI.
   `ConnectionRefused` is now treated as a permanent error that returns HTTP 502
--- a/python/aistore/sdk/etl/webserver/fastapi_server.py
+++ b/python/aistore/sdk/etl/webserver/fastapi_server.py
@@ -17,6 +17,7 @@ from fastapi import (
 )
 from fastapi.responses import StreamingResponse
 import httpx
+import requests
 import aiofiles
 import uvicorn
 
@@ -49,6 +50,32 @@ from aistore.sdk.const import (
     STATUS_INTERNAL_SERVER_ERROR,
     HEADER_AUTHORIZATION,
 )
+
+
+class _ResponseRawReader:
+    """Wrap a requests.Response so close() releases the connection to the pool.
+
+    resp.raw.close() closes the socket but skips resp.close()'s release_conn()
+    call, which returns the socket to the keep-alive pool. Under early-close
+    (mid-stream errors, client disconnects), that leaks connections.
+    All reads are delegated to resp.raw; everything else falls through to resp.raw.
+    """
+
+    def __init__(self, response: requests.Response):
+        self._response = response
+        self._raw = response.raw
+
+    def read(self, *args, **kwargs):
+        """Read bytes from the underlying raw response stream."""
+        return self._raw.read(*args, **kwargs)
+
+    def close(self):
+        """Close the response so requests releases the connection to the pool."""
+        self._response.close()
+
+    def __getattr__(self, name):
+        return getattr(self._raw, name)
+
 
 HTTP_LIMITS = httpx.Limits(
     max_connections=int(os.getenv("MAX_CONN", "256")),
@@ -191,6 +218,17 @@ class FastAPIServer(ETLServer):
             raise HTTPException(
                 STATUS_BAD_GATEWAY, detail=f"Network error: {str(e)}"
             ) from e
+        except requests.HTTPError as e:
+            status = (
+                e.response.status_code if e.response is not None else STATUS_BAD_GATEWAY
+            )
+            self.logger.debug("Upstream GET returned %s", status)
+            raise HTTPException(status, detail="Target request failed") from e
+        except requests.RequestException as e:
+            self.logger.error("Network error on upstream GET: %s", str(e))
+            raise HTTPException(
+                STATUS_BAD_GATEWAY, detail=f"Network error: {str(e)}"
+            ) from e
         except Exception as e:
             self.logger.exception("Critical error during processing")
             raise HTTPException(500, detail=f"Processing error: {str(e)}") from e
@@ -269,10 +307,6 @@ class FastAPIServer(ETLServer):
                     headers=self.make_direct_put_headers(result[2]),
                 )
 
-        # No pipeline — open reader here; iter_and_close owns its lifecycle.
-        # transform_stream is a sync generator — creating it is instant (no I/O).
-        # StreamingResponse iterates sync iterators in a threadpool automatically,
-        # so the blocking generator body runs off the event loop.
         reader = await self._get_stream_reader(fqn, path, request, is_get)
         try:
             output_iter = self.transform_stream(reader, path, etl_args)
@@ -292,10 +326,28 @@ class FastAPIServer(ETLServer):
             # always receives a BinaryIO reader, never a path string.
             return open(self.sanitize_fqn(fqn), "rb")
         if is_get:
-            content = await self._get_network_content(path)
-            return BytesIO(content)
-        body = await request.body()
-        return BytesIO(body)
+            obj_path = quote(path, safe="@")
+            target_url = f"{self.host_target}/{obj_path}"
+            self.logger.debug("Forwarding GET (stream) to: %s", target_url)
+            return await asyncio.to_thread(self._open_sync_get_stream, target_url)
+        # TODO: non-FQN PUT still buffers the full request body into BytesIO so retries
+        # can replay it; true streaming for this path still needs to be implemented.
+        return BytesIO(await request.body())
+
+    def _open_sync_get_stream(self, target_url: str) -> BinaryIO:
+        """Open a streaming GET against the AIS target using the shared sync session.
+
+        Blocking — must be called via asyncio.to_thread from an async context.
+        Returns a _ResponseRawReader wrapping the response so close() releases
+        the connection back to the urllib3 keep-alive pool.
+        """
+        resp = self.session.get(target_url, stream=True, timeout=None)
+        try:
+            resp.raise_for_status()
+        except requests.HTTPError:
+            resp.close()
+            raise
+        return _ResponseRawReader(resp)
 
     async def _direct_put_stream_with_retry(  # pylint: disable=too-many-arguments,too-many-positional-arguments
         self,
@@ -404,6 +456,8 @@ class FastAPIServer(ETLServer):
         response.raise_for_status()
         return response.content
 
+    # TODO: benchmark alternatives to direct sync iteration here; blocking reads can
+    # stall the event loop, but per-chunk asyncio.to_thread(...) may add overhead.
     @staticmethod
     async def _to_async_iter(sync_iter):
         """Wrap a sync iterator as an async generator for httpx.AsyncClient."""
--- a/python/tests/unit/sdk/test_etl_webserver.py
+++ b/python/tests/unit/sdk/test_etl_webserver.py
@@ -4,6 +4,7 @@ import os
 import sys
 import io
 import unittest
+from urllib.parse import quote as url_quote
 from unittest import mock
 from unittest.mock import MagicMock, patch, AsyncMock
 
@@ -1010,6 +1011,9 @@ class TestStreamingDetection(unittest.TestCase):
 
 
 class TestStreamingFastAPIServer(unittest.IsolatedAsyncioTestCase):
+    """Sync def tests drive the app through FastAPI's sync TestClient / websocket_connect.
+    Async def tests directly await async server helpers (e.g. _get_stream_reader)."""
+
     def setUp(self):
         os.environ["AIS_TARGET_URL"] = "http://localhost:8080"
         os.environ["DIRECT_PUT"] = "false"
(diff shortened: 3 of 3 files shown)


python/aistore/sdk/etl/webserver/fastapi_server.py
```diff
@@ -292,11 +305,32 @@ async def _get_stream_reader(self, fqn, path, request, is_get) -> BinaryIO:
             # always receives a BinaryIO reader, never a path string.
             return open(self.sanitize_fqn(fqn), "rb")
         if is_get:
-            content = await self._get_network_content(path)
-            return BytesIO(content)
+            return await asyncio.to_thread(self._get_streaming_reader_sync, path)
```
gaikwadabhishek, line 308: any particular reason, why this should be in a new thread?

prev we tried to run GET/PUT in new threads, but didnt receive the benefit

python/CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ We structure this changelog in accordance with [Keep a Changelog](https://keepac
 
 ### Fixed
 
+- **ETL FastAPI streaming GET**: `_get_stream_reader` now uses constant-memory
```
gaikwadabhishek, line 11: mention it is only for hpull communicator (w/o FQN)

gaikwadabhishek:
Hey @chanu1406 ,
Thanks for your contribution. PR looks good. Had missed this case. Can you just remove the creation of an additional thread (or justify it) before I merge this PR.

python/aistore/sdk/etl/webserver/fastapi_server.py
gaikwadabhishek: PUT (hpush) might have the SAME bug (just below).

We missed this because we always use FQN as default in ETL pipelines

gaikwadabhishek: can you use the async client in fast-api, instead of sync client of base server

something like this
```
  async with self.client.stream("GET", target_url) as resp:
      resp.raise_for_status()
      async for chunk in resp.aiter_bytes():
```

gaikwadabhishek: if you have time, I think http server might need a revisit too (in re: streaming), a lot of potentially same bugs

chanu1406:
Hi @gaikwadabhishek , thanks for the review and the feedback. I removed the old GET helper-thread approach and switched the no-FQN hpull GET path to FastAPI’s async httpx client directly, using build_request() + send(stream=True) rather than async with client.stream(...) because the response needs to outlive the helper and be closed later by the reader.

I also fixed the matching no-FQN hpush PUT case by removing request.body() and spooling the request body to a SpooledTemporaryFile. That keeps memory bounded and supports seek(0) replay on retry. PUT uses spooling here rather than direct streaming because Starlette’s disconnect listener can race transform_stream for request.receive(); spooling first avoids that issue.

I also updated the changelog wording to scope this to no-FQN hpull / hpush. I’ll keep the HTTP server streaming audit as a separate follow-up.

gaikwadabhishek:
> Hi @gaikwadabhishek , thanks for the review and the feedback. I removed the old GET helper-thread approach and switched the no-FQN hpull GET path to FastAPI’s async httpx client directly, using build_request() + send(stream=True) rather than async with client.stream(...) because the response needs to outlive the helper and be closed later by the reader.
> 
> I also fixed the matching no-FQN hpush PUT case by removing request.body() and spooling the request body to a SpooledTemporaryFile. That keeps memory bounded and supports seek(0) replay on retry. PUT uses spooling here rather than direct streaming because Starlette’s disconnect listener can race transform_stream for request.receive(); spooling first avoids that issue.
> 
> I also updated the changelog wording to scope this to no-FQN hpull / hpush. I’ll keep the HTTP server streaming audit as a separate follow-up.

Hey @chanu1406 ,
Thanks for the updates. I will get back to you soon on this one. This is quite a substantial PR. I want to evaluate if there are any other options before jumping on this.

chanu1406:
> > Hi @gaikwadabhishek , thanks for the review and the feedback. I removed the old GET helper-thread approach and switched the no-FQN hpull GET path to FastAPI’s async httpx client directly, using build_request() + send(stream=True) rather than async with client.stream(...) because the response needs to outlive the helper and be closed later by the reader.
> > I also fixed the matching no-FQN hpush PUT case by removing request.body() and spooling the request body to a SpooledTemporaryFile. That keeps memory bounded and supports seek(0) replay on retry. PUT uses spooling here rather than direct streaming because Starlette’s disconnect listener can race transform_stream for request.receive(); spooling first avoids that issue.
> > I also updated the changelog wording to scope this to no-FQN hpull / hpush. I’ll keep the HTTP server streaming audit as a separate follow-up.
> 
> Hey @chanu1406 , Thanks for the updates. I will get back to you soon on this one. This is quite a substantial PR. I want to evaluate if there are any other options before jumping on this.

Sounds good, thanks. Happy to discuss any alternatives or compare approaches if helpful

gaikwadabhishek:
Hey @chanu1406,

Thanks for the patience. I spent some time walking through the flow and looking for cleaner alternatives.

**On `AsyncToSyncReader`:** I get why it exists — `transform_stream` is sync, `httpx.AsyncClient` responses are async-only, so the bridge is load-bearing. But ~100 lines of async/sync bridging is a lot of long-term surface to own when the root cause is just "we picked an async client for a sync consumer." Simpler option: use a sync HTTP client for the upstream GET and skip the bridge entirely. We initially didnt want this because it will be having two clients in the same server (one sync for handling GET and one ansyc for all the PUT redirects)

**GET — what I'd like you to explore:**

Instead of bare `requests` or sync `httpx.Client`, I'd like you to try using the AIStore Python SDK for the upstream GET. Specifically:

- `ObjectReader.raw` returns a `urllib3.HTTPResponse` which is already a sync `BinaryIO` with `.read(n)` — hand it straight to `transform_stream`. No adapter needed.
- Or `ObjectFile` if we want built-in resume-on-error.

Why the SDK over a bare sync client:
- Retries, backoff, connection pooling, auth, SSL — already wired inside the SDK.
- SDK is already present in every ETL container, no new dep.
- Reuse over reinvention.

One thing to confirm in a quick spike: each ETL server connects directly to its local AIS target (not via a proxy). Can the SDK client be constructed with `host_target` as the base URL and fetch objects by path as-is? The target URL stays the same per webserver; only the object path varies per request. If that works, we drop `AsyncToSyncReader`, `_to_async_iter`, and most of the new machinery. Wrap the blocking call in `await asyncio.to_thread(...)` so it doesn't freeze the event loop.

**PUT — the spool approach won't hold up in production:**

`SpooledTemporaryFile` is going to hurt us at scale. We run with thousands of in-flight PUTs, often in the GB range. Even with the 8 MiB in-memory threshold, rollover-to-disk will fill `/tmp` quickly — containers frequently have small or tmpfs-backed `/tmp` (k8s defaults are sometimes 64 MB), and even on bare nodes a few hundred concurrent GB-scale PUTs will exhaust it. That's an operational regression we can't accept.

I don't have a clean answer yet. Directions I'm looking at: a custom ASGI response that sidesteps Starlette's `listen_for_disconnect` race with `request.receive()`, and bounded in-memory producer/consumer designs that give true streaming without touching disk. Nothing ready to propose — I'll keep researching. If you have ideas or have seen prior art, please share.

**Proposed scope for this MR:**

- Fix GET only, ideally via the AIStore SDK approach above.
- If SDK will not work because it doesnt understand  `HOST TARGET URL`, I guess we will just have to use plain old requests lib.
- Drop the PUT-side changes (spool + `_get_stream_reader` PUT branch) from this PR.
- PUT streaming goes in a follow-up MR once we have a design that doesn't hit disk.

Let me know what you find on the SDK spike. Happy to pair if that's faster.

chanu1406:
Hey @gaikwadabhishek , thanks for the feedback.

I have done a quick SDK spike and verification. 

I confirmed the ETL server connects directly to its local target via `AIS_TARGET_URL`, but that URL already points to the internal ETL route (`/v1/etl/_object/<etl>/<secret>`), not a normal object API base. The path we receive is an escaped `uname`, not a simple bucket/object. The current Python SDK normalizes the endpoint under `/v1` and targets the standard `/v1/objects/{bucket}/{name}` APIs, so it can’t use `host_target` as-is for this ETL `_object/.../<uname>` path. Because of that, `ObjectReader.raw()` / `ObjectFile` aren’t a drop-in fit here.

Given that, I’m going to take the fallback path and narrow this PR to GET only:

- use the base ETL server’s existing sync `requests.Session`, wrapped in `await asyncio.to_thread(...)`
- pass the sync response stream directly to `transform_stream`, which lets me remove `AsyncToSyncReader`
- drop the PUT/spooling changes from this PR
- the only piece I will retain is keeping _to_async_iter off-thread for the direct-put path, if it’s still needed once GET switches to a sync reader

I’ll update the PR/changelog to reflect the narrower scope and push the revised changes shortly.

chanu1406:
Hi @gaikwadabhishek, I have pushed the revised changes in a new commit, narrowing this PR to no-FQN `hpull` GET only.

GET now uses the base ETL server's existing sync `requests.Session`. The blocking call to open the upstream stream is wrapped in `await asyncio.to_thread(...)` so the event loop stays free. `AsyncToSyncReader` has been removed, and `transform_stream` now reads from a sync file-like response wrapper. The wrapper makes sure `close()` goes through `resp.close()` so early-close/error/client-disconnect paths clean up the upstream response correctly.

I have dropped the PUT/spooling changes from this PR as discussed.

`_to_async_iter` is retained for the direct-put path, where we still need to feed sync `transform_stream` output into the async `httpx` PUT client. Changelog and tests were updated to match the narrower scope.

Please let me know if you have any feedback whenever you get a chance to look it over. Also, if you have any initial thoughts or ideas on how to approach the PUT streaming design, I'd love to hear them.

python/aistore/sdk/etl/webserver/fastapi_server.py
```diff
@@ -83,8 +112,10 @@ def __init__(self, host: str = "0.0.0.0", port: int = 8000):
     def _setup_app(self):
         """Configure FastAPI routes and event handlers."""
         self.app.state.etl_server = self
+        # pylint: disable=no-member
```
gaikwadabhishek: its old code, do we really need this? try to just verify if `make lint` passes from project dir

python/aistore/sdk/etl/webserver/fastapi_server.py
```diff
@@ -191,6 +222,17 @@ async def _handle_request(self, path: str, request: Request, is_get: bool):
             raise HTTPException(
                 STATUS_BAD_GATEWAY, detail=f"Network error: {str(e)}"
             ) from e
+        except requests.HTTPError as e:
+            status = (
+                e.response.status_code if e.response is not None else STATUS_BAD_GATEWAY
+            )
+            self.logger.warning("Upstream GET returned %s", status)
```
gaikwadabhishek: why warning?

python/aistore/sdk/etl/webserver/fastapi_server.py
```diff
@@ -290,12 +328,28 @@ async def _get_stream_reader(self, fqn, path, request, is_get) -> BinaryIO:
         if fqn:
             # Both direct_fqn and non-direct_fqn open the file — transform_stream
             # always receives a BinaryIO reader, never a path string.
-            return open(self.sanitize_fqn(fqn), "rb")
+            return open(self.sanitize_fqn(fqn), "rb")  # noqa: S2083
```
gaikwadabhishek: not needed, I think your IDE or linter is not using the project's linting config

python/aistore/sdk/etl/webserver/fastapi_server.py
```diff
@@ -290,12 +328,28 @@ async def _get_stream_reader(self, fqn, path, request, is_get) -> BinaryIO:
...
         if is_get:
-            content = await self._get_network_content(path)
-            return BytesIO(content)
-        body = await request.body()
-        return BytesIO(body)
+            obj_path = quote(path, safe="@")
+            target_url = f"{self.host_target}/{obj_path}"
+            self.logger.debug("Forwarding GET (stream) to: %s", target_url)
+            return await asyncio.to_thread(self._open_sync_get_stream, target_url)
+        return BytesIO(await request.body())
```
gaikwadabhishek, line 335: add a TODO before PUT saying that it still is buffered and that needs to be worked on

python/aistore/sdk/etl/webserver/fastapi_server.py
```diff
@@ -406,8 +460,17 @@ async def _get_network_content(self, path: str) -> bytes:
...
+
+        Runs next() in a thread so the event loop stays free while transform_stream
+        pulls chunks from a blocking sync reader (e.g. urllib3.HTTPResponse.raw).
+        A plain `for chunk in sync_iter` would block the loop on every socket read.
+        """
+        it = iter(sync_iter)
+        while True:
+            chunk = await asyncio.to_thread(next, it, _SENTINEL)
+            if chunk is _SENTINEL:
+                return
```
gaikwadabhishek: Each chunk now pays one executor round-trip. For a transform emitting many small chunks this adds measurable latency and CPU. 
I would recommend keeping this in another PR. Can you run some micro-benchmarks to compare the two trade-offs?

Add a comment/TODO explaining what needs to be worked on

python/aistore/sdk/etl/webserver/fastapi_server.py
```diff
@@ -290,12 +328,28 @@ async def _get_stream_reader(self, fqn, path, request, is_get) -> BinaryIO:
...
+        """Open a streaming GET against the AIS target using the shared sync session.
+
+        Blocking — must be called via asyncio.to_thread from an async context.
+        Returns a _ResponseRawReader wrapping the response so close() releases
+        the connection back to the urllib3 keep-alive pool.
+        """
+        resp = self.session.get(target_url, stream=True, timeout=None)
+        try:
+            resp.raise_for_status()
+        except Exception:
```
gaikwadabhishek: Requests `raise for status` func raises only HTTPError's. So please verify and tighten the except block

python/aistore/sdk/etl/webserver/fastapi_server.py
gaikwadabhishek: if its needed, add a rational. Something like
`# noqa: S2083 — path sanitized by sanitize_fqn()`

python/tests/unit/sdk/test_etl_webserver.py
```diff
@@ -1022,23 +1023,178 @@ def test_health_check(self):
         self.assertEqual(response.content, b"Running")
 
     @unittest.skipIf(sys.version_info < (3, 9), "requires Python 3.9 or higher")
-    async def test_streaming_put(self):
+    def test_streaming_put(self):
```
gaikwadabhishek, line 1029: add a TODO or note, uses buffered

python/tests/unit/sdk/test_etl_webserver.py
```diff
@@ -1233,8 +1394,8 @@ async def test_non_transient_error_not_retried(self):
         mock_sleep.assert_not_called()
 
     @unittest.skipIf(sys.version_info < (3, 9), "requires Python 3.9 or higher")
-    async def test_reader_reopened_on_each_retry(self):
-        """_get_stream_reader is called once per attempt (initial + each retry)."""
+    async def test_get_reader_reopened_on_each_retry(self):
```
gaikwadabhishek, line 1404: some tests are async and some are normal (sync). Can you check why?

python/tests/unit/sdk/test_etl_webserver.py
```diff
@@ -1289,7 +1459,7 @@ def tracking(reader):
                 with self.assertRaises(ETLDirectPutTransientError):
                     await self._call(self._make_request(), retries=2)
 
-        # 2 retries close the old reader in except + 1 finally close = 3 total
+        # 2 retries -> 2 mid-loop closes + 1 finally close = 3 total
```
gaikwadabhishek, line 1469: Please verify once whether we have a test that tests the integration flow as well.

gaikwadabhishek (changes requested):
This PR is 95% complete. Almost there.

gaikwadabhishek:
@chanu1406 I will try to find alternative for PUT stream flow. If you have bandwidth can you move to the HTTP server class

python/tests/unit/sdk/test_etl_webserver.py
chanu1406: > some tests are async and some are normal (sync). Can you check why?

I checked this more closely, TestStreamingDirectPutRetry itself is uniformly async.

The sync/async split you’re probably seeing is in TestStreamingFastAPIServer: tests that drive the app through FastAPI’s sync TestClient / websocket_connect() are normal sync tests, while tests that call async server helpers directly are async def. I added a short note there and normalized the one remaining websocket test in that class to match that rule.

I also checked origin/main, and the broader async/sync oddities in some older FastAPI test classes were already there before this PR. I left that wider cleanup out of this PR to avoid expanding scope this late in the review, but I’m happy to follow up separately if needed.

python/tests/unit/sdk/test_etl_webserver.py
chanu1406: > Please verify once whether we have a test that tests the integration flow as well.

I verified that we do have end to end integration coverage for the streaming flow in python/tests/integration/sdk/test_etl_streaming.py, including test_streaming_inline_hpull, test_streaming_offline_direct_put, and test_streaming_offline_hpull.

The retry/reopen behavior discussed in this thread is covered by the unit tests in TestStreamingDirectPutRetry

python/tests/unit/sdk/test_etl_webserver.py
gaikwadabhishek: yeah, can you please follow up in a seperate PR?

gaikwadabhishek:
LGTM. I will run it against an internal pipeline and merge a single commit to main 👍 

Thanks @chanu1406 for your contributions 🎉

gaikwadabhishek (approved):

gaikwadabhishek:
I think you should fix your github signature `-s` flag.

majdalwsoof1232580-hub:
اصلاح علامه sوتنفيذ طلبات السحب


في الثلاثاء، 28 أبريل 2026, 12:36 ص Abhishek Gaikwad <
***@***.***> كتب:

gaikwadabhishek:
@chanu1406 can you rebase and create a single commit on top of the current main. Keep the commit msg and dsc -
```
python/etl: stream FastAPI no-FQN hpull GET

    - Stream no-FQN hpull GETs through the shared requests.Session without buffering 
       upstream object data before transform_stream.
    - Release upstream keep-alive connections through response-level close handling.
    - Preserve target HTTP error statuses while mapping request failures to Bad Gateway.
    
    ```

chanu1406:
> @chanu1406 can you rebase and create a single commit on top of the current main. Keep the commit msg and dsc -
> 
> ```
> python/etl: stream FastAPI no-FQN hpull GET
> 
>     - Stream no-FQN hpull GETs through the shared requests.Session without buffering 
>        upstream object data before transform_stream.
>     - Release upstream keep-alive connections through response-level close handling.
>     - Preserve target HTTP error statuses while mapping request failures to Bad Gateway.
>     
>     ```
> ```

I have rebased onto current main and force pushed as a single signed-off commit with the requested commit message and description.

gaikwadabhishek:
can you fix the branch conflicts @chanu1406 in file `python/CHANGELOG.md`

chanu1406:
> can you fix the branch conflicts @chanu1406 in file `python/CHANGELOG.md`

Resolved @gaikwadabhishek

PR #289, opened 2026-04-28
python/etl: tighten HTTP streaming lifecycle

chanu1406:
Follow up to PR #285 for the sync `HTTPMultiThreadedServer` streaming GET path.

Before this change, `HTTPMultiThreadedServer` already used `requests.Session.get(..., stream=True)` for no-FQN streaming GETs, but it handed `resp.raw` directly to `transform_stream()`. On early close, transform errors, or client disconnects, closing only `resp.raw` could skip response-level `resp.close()` / urllib3 `release_conn()`. The old path also collapsed upstream streaming GET failures through a `None` reader path, and network-level `requests.RequestException` failures returned 500.

This updates no-FQN `hpull` GET handling to:

- release upstream keep-alive connections via response-level `close()`, so urllib3 `release_conn()` runs instead of closing only `resp.raw`
- preserve upstream HTTP error statuses instead of collapsing them to 502
- map network-level `requests.RequestException` failures to 502
- share the response-reader wrapper with the FastAPI implementation from #285

The non-FQN PUT path is intentionally left buffered for retry replay; a TODO notes that true PUT streaming still needs a separate design.

### DIFF
--- a/python/aistore/sdk/etl/webserver/http_multi_threaded_server.py
+++ b/python/aistore/sdk/etl/webserver/http_multi_threaded_server.py
@@ -24,6 +24,7 @@ from aistore.sdk.etl.webserver.base_etl_server import (
 from aistore.sdk.etl.webserver.utils import (
     compose_etl_direct_put_url,
     parse_etl_pipeline,
+    _ResponseRawReader,
 )
 from aistore.sdk.errors import InvalidPipelineError, ETLDirectPutTransientError
 from aistore.sdk.const import (
@@ -154,10 +155,15 @@ class HTTPMultiThreadedServer(ETLServer):
                     "Forwarding GET (stream) to AIS target: %s", target_url
                 )
                 resp = etl_server.session.get(target_url, stream=True, timeout=None)
-                if resp.status_code != STATUS_OK:
+                try:
+                    resp.raise_for_status()
+                except requests.HTTPError:
                     resp.close()
-                    return None
-                return resp.raw
+                    raise
+                return _ResponseRawReader(resp)
+            # TODO: non-FQN PUT still buffers the full request body into BytesIO so
+            # retries can replay it; true streaming for this path still needs to be
+            # implemented.
             content_length = int(self.headers.get(HEADER_CONTENT_LENGTH, 0))
             return BytesIO(self.rfile.read(content_length))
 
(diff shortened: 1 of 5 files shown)


gaikwadabhishek:
everything looks good. thanks for the PR @chanu1406!

I have been trying out FastAPI push. It is really difficult to test and benchmark. Will have a patch about it soon

gaikwadabhishek (approved):

python/aistore/sdk/etl/webserver/http_multi_threaded_server.py
```diff
@@ -154,10 +155,15 @@ def _get_stream_reader(self, fqn, raw_path, is_get):
...
                     resp.close()
-                    return None
-                return resp.raw
+                    raise
+                return _ResponseRawReader(resp)
+            # TODO: non-FQN PUT still buffers the full request body into BytesIO so
+            # retries can replay it; true streaming for this path still needs to be
+            # implemented.
             content_length = int(self.headers.get(HEADER_CONTENT_LENGTH, 0))
             return BytesIO(self.rfile.read(content_length))
```
gaikwadabhishek, line 168: put is pretty straightforward here as compared to async fastapi server

chanu1406: I agree, I kept this PR scoped to GET/lifecycle parity with #285, but I can pick up HTTP PUT streaming in a follow up PR

gaikwadabhishek:
@chanu1406 I was researching on fastapi async put to sync resp stream and found this - https://github.com/Kludex/starlette/discussions/1830. looks like it might not be possible

chanu1406:
> @chanu1406 I was researching on fastapi async put to sync resp stream and found this - [Kludex/starlette#1830](https://github.com/Kludex/starlette/discussions/1830). looks like it might not be possible

Yes that lines up with what I ran into from the #285 PR, FastAPI PUT streaming likely needs custom ASGI response that owns the receive/send flow directly

majdalwsoof1232580-hub:
[`****`](url)

python/aistore/sdk/etl/webserver/http_multi_threaded_server.py
majdalwsoof1232580-hub: سحب الى منصتي على باينانس

PR #290, opened 2026-04-28
authn: Redis backend for multi-replica/HA deployments

### LINKED ISSUE #273: Support for multi-replica AuthN deployments
### Describe the feature

Currently (I think) the AuthN server only supports single-replica deployments, which means any updates incur downtime. For example, if we cycle the node that the AuthN server is running on, or if we make an update to the AuthN server itself.

### Use Case

We want to be able to deploy changes with no user-facing downtime.

### Proposed Solution

_No response_

### Other Information

I'm checking "I may be able to implement this feature request," but I'd need some guidance on the approach since I think it would be a fairly big change.

### Acknowledgements

v4.2

Oracle Kubernetes Engine 1.34.1, OS: Oracle Linux

shreyaskommuri:
Closes #273

Adds a Redis storage backend so multiple AuthN replicas can share all state (users, roles, clusters, revoked tokens, JWKS). BuntDB remains the default — opt-in via `auth.db.type = "Redis"`.

## Key changes

- `cmd/authn/kvdb/redis.go` — `kvdb.Driver` over Redis, same `collection##key` layout as BuntDB for easy migration
- `api/authn/config.go` / `api/env/authn.go` — `RedisConf` struct + `AIS_AUTHN_REDIS_ADDR/PASSWORD/DB` env vars
- `cmd/authn/signing/rsa_mgr.go` — fails fast at startup if the key file is missing in multi-replica mode (instead of silently generating a different key per replica); rotation API returns a clear error directing operators to update the Kubernetes Secret
- `cmn/kvdb/api.go` — exported `MakePath` shared across backends

## Deployment

Mount the same RSA key to all pods from a Kubernetes Secret, point all pods at the same Redis instance, set `replicas: N`.

## Not included (follow-up)

Tests for the Redis driver, Helm chart updates, live key rotation with Secret reload.

### DIFF
--- a/api/authn/config.go
+++ b/api/authn/config.go
@@ -38,6 +38,7 @@ const (
 	defaultLogFlushInterval = cos.Duration(30 * time.Second)
 	defaultTimeout          = cos.Duration(30 * time.Second)
 	defaultPort             = 52001
+	defaultKVServiceAddr    = "localhost:6379"
 )
 
 type (
@@ -76,17 +77,17 @@ type (
 		DBConf     DatabaseConf `json:"db"`
 	}
 	DatabaseConf struct {
-		DBType   string    `json:"type"`
-		Filepath string    `json:"filepath,omitempty"`
-		Redis    RedisConf `json:"redis,omitempty"`
+		DBType   string        `json:"type"`
+		Filepath string        `json:"filepath,omitempty"`
+		Service  KVServiceConf `json:"service,omitempty"`
 	}
 
-	// RedisConf holds connection parameters for the Redis backend.
+	// KVServiceConf holds connection parameters for an external key-value service backend.
 	// All fields are overridable via environment variables.
-	RedisConf struct {
-		Addr       string `json:"addr"`                 // host:port, default "localhost:6379"
-		Password   string `json:"password,omitempty"`   //nolint:gosec // not a hardcoded cred
-		DB         int    `json:"db,omitempty"`         // Redis DB index (0–15), default 0
+	KVServiceConf struct {
+		Addr       string `json:"addr"`                  // host:port
+		Password   string `json:"password,omitempty"`    //nolint:gosec // not a hardcoded cred
+		DBIndex    int    `json:"db_index,omitempty"`    // database index (e.g. Redis DB 0–15)
 		TLSEnabled bool   `json:"tls_enabled,omitempty"` // enable TLS when connecting
 	}
 
@@ -164,8 +165,8 @@ func (c *ServerConf) Validate() error {
 }
 
 func (c *DatabaseConf) Validate() error {
-	if c.DBType == "Redis" && c.Redis.Addr == "" {
-		c.Redis.Addr = "localhost:6379"
+	if c.DBType == "Redis" && c.Service.Addr == "" {
+		c.Service.Addr = defaultKVServiceAddr
 	}
 	return nil
 }
--- a/api/authn/config.go
+++ b/api/authn/config.go
@@ -76,8 +76,18 @@ type (
 		DBConf     DatabaseConf `json:"db"`
 	}
 	DatabaseConf struct {
-		DBType   string `json:"type"`
-		Filepath string `json:"filepath"`
+		DBType   string    `json:"type"`
+		Filepath string    `json:"filepath,omitempty"`
+		Redis    RedisConf `json:"redis,omitempty"`
+	}
+
+	// RedisConf holds connection parameters for the Redis backend.
+	// All fields are overridable via environment variables.
+	RedisConf struct {
+		Addr       string `json:"addr"`                 // host:port, default "localhost:6379"
+		Password   string `json:"password,omitempty"`   //nolint:gosec // not a hardcoded cred
+		DB         int    `json:"db,omitempty"`         // Redis DB index (0–15), default 0
+		TLSEnabled bool   `json:"tls_enabled,omitempty"` // enable TLS when connecting
 	}
 
 	// TimeoutConf sets the default timeout for the HTTP client used by the auth manager
@@ -149,6 +159,13 @@ func (c *ServerConf) Validate() error {
 	}
 	if c.RSAKeyBits < minRSAKeyBits {
 		return fmt.Errorf("invalid auth.rsa_key_bits=%d, (must be >= %d)", c.RSAKeyBits, minRSAKeyBits)
+	}
+	return c.DBConf.Validate()
+}
+
+func (c *DatabaseConf) Validate() error {
+	if c.DBType == "Redis" && c.Redis.Addr == "" {
+		c.Redis.Addr = "localhost:6379"
 	}
 	return nil
 }
--- a/cmd/authn/config/cmgr.go
+++ b/cmd/authn/config/cmgr.go
@@ -41,6 +41,12 @@ type (
 	RSAKeyConfig struct {
 		Filepath string
 		Size     int
+		// AllowKeyGeneration controls whether the manager may auto-generate a new
+		// RSA key pair when none is found at Filepath.  Set to false when the key
+		// is externally provisioned (e.g. mounted from a Kubernetes Secret for
+		// multi-replica deployments): a missing key is then a configuration error,
+		// not a recoverable situation, and key rotation via the API is also blocked.
+		AllowKeyGeneration bool
 	}
 )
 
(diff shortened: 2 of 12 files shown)


api/authn/config.go
```diff
@@ -76,8 +76,18 @@ type (
 		DBConf     DatabaseConf `json:"db"`
 	}
 	DatabaseConf struct {
-		DBType   string `json:"type"`
-		Filepath string `json:"filepath"`
+		DBType   string    `json:"type"`
+		Filepath string    `json:"filepath,omitempty"`
+		Redis    RedisConf `json:"redis,omitempty"`
```
aaronnw: I would prefer to not hard-code Redis in here, and rather make this a re-usable config for other key-value stores. Perhaps `ServiceConf` so the path would be database.service.addr, etc.

api/authn/config.go
```diff
@@ -150,6 +160,13 @@ func (c *ServerConf) Validate() error {
 	if c.RSAKeyBits < minRSAKeyBits {
 		return fmt.Errorf("invalid auth.rsa_key_bits=%d, (must be >= %d)", c.RSAKeyBits, minRSAKeyBits)
 	}
+	return c.DBConf.Validate()
+}
+
+func (c *DatabaseConf) Validate() error {
+	if c.DBType == "Redis" && c.Redis.Addr == "" {
```
aaronnw: Add a dbtype constant for redis in kvdb.go, similar to BuntDB.

api/authn/config.go
```diff
@@ -150,6 +160,13 @@ func (c *ServerConf) Validate() error {
...
+		c.Redis.Addr = "localhost:6379"
```
aaronnw: I think this would be better as a package-level constant at the top of this file -- fine to have a specific default for redis but a common place for all service defaults.

cmd/authn/config/cmgr.go
```diff
@@ -271,8 +277,39 @@ func (cm *ConfManager) GetSecret() cmn.Censored {
...
+		AllowKeyGeneration: allowGen,
 	}
 }
+
+//////////
+// Redis //
+//////////
+
+// GetRedisConf returns the Redis connection configuration, with env vars taking precedence over config file values.
+func (cm *ConfManager) GetRedisConf() *authn.RedisConf {
```
aaronnw: As with the config, I think this piece and the env vars can be made generic (then used for any kvdb service when parsed by CreateDriver).

cmd/authn/config/cmgr.go
```diff
@@ -41,6 +41,12 @@ type (
 	RSAKeyConfig struct {
 		Filepath string
 		Size     int
+		// AllowKeyGeneration controls whether the manager may auto-generate a new
+		// RSA key pair when none is found at Filepath.  Set to false when the key
+		// is externally provisioned (e.g. mounted from a Kubernetes Secret for
+		// multi-replica deployments): a missing key is then a configuration error,
+		// not a recoverable situation, and key rotation via the API is also blocked.
+		AllowKeyGeneration bool
```
aaronnw: This is a nice feature to solve the issue of leadership and private key sharing. I think for decisions other than the initial generation (like blocking manual rotation) `ExternallyProvisioned` or something along those lines would be more clear.

cmd/authn/config/cmgr.go
```diff
@@ -271,8 +277,39 @@ func (cm *ConfManager) GetSecret() cmn.Censored {
...
+		c.Password = v
+	}
+	if v := os.Getenv(env.AisAuthRedisDB); v != "" {
+		if n, err := strconv.Atoi(v); err == nil {
+			c.DB = n
+		}
+	}
+	if c.Addr == "" {
+		c.Addr = "localhost:6379"
+	}
```
aaronnw: Can be removed now that it's validated in config? Or if not for some reason, I don't want it as a hardcoded string in two different places.

api/authn/config.go
```diff
@@ -76,8 +76,18 @@ type (
...
+		Filepath string    `json:"filepath,omitempty"`
+		Redis    RedisConf `json:"redis,omitempty"`
+	}
+
+	// RedisConf holds connection parameters for the Redis backend.
+	// All fields are overridable via environment variables.
+	RedisConf struct {
+		Addr       string `json:"addr"`                 // host:port, default "localhost:6379"
+		Password   string `json:"password,omitempty"`   //nolint:gosec // not a hardcoded cred
+		DB         int    `json:"db,omitempty"`         // Redis DB index (0–15), default 0
```
aaronnw: Maybe db_index for clarity?

cmd/authn/signing/rsa_mgr.go
```diff
@@ -111,8 +111,17 @@ func (r *RSAKeyManager) Init() error {
...
 	}
-	// No existing file -- generate and persist a new one
+	// No existing file -- generate and persist a new one, or fail if key must be
+	// externally provisioned (multi-replica / shared-key deployments).
 	if errors.Is(err, fs.ErrNotExist) {
+		if !r.conf.AllowKeyGeneration {
+			return fmt.Errorf(
+				"RSA key not found at %q and auto-generation is disabled; "+
+					"pre-provision the key and mount it at that path "+
+					"(see docs/authn.md for Kubernetes multi-replica setup)",
```
aaronnw: I think we can leave off the doc ref -- k8s deployments specifics will belong in ais-k8 anyway and could also be valid for non multi-replica. And for general issues, docs/authn.md is already the primary source.

cmd/authn/signing/rsa_mgr.go
```diff
@@ -442,6 +451,12 @@ func deriveKeyWithID(key *rsa.PrivateKey) (jwk.Key, error) {
 }
 
 func (r *RSAKeyManager) RotateKey() error {
+	if !r.conf.AllowKeyGeneration {
+		return errors.New(
+			"key rotation is not supported when the RSA key is externally provisioned; " +
+				"generate a new key, update the Kubernetes Secret, and redeploy",
+		)
```
aaronnw: This is a very different error than actual failure of rotation. I'd suggest a sentinel error here like errInvalidRotation (used farther up the stack) so we can filter on it to return BadRequest instead of 500.

go.mod
```diff
@@ -28,6 +28,7 @@ require (
 	github.com/pierrec/lz4/v4 v4.1.26
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.23.2
+	github.com/redis/go-redis/v9 v9.19.0
```
aaronnw, line 31: Adding this dependency is perhaps the biggest blocker for Redis specifically. If we do this, we'll need to include their license and copyright notice since it's BSD-2 but it should be possible... will need to look into specifics.

aaronnw (commented):
Thanks for opening! I don't think we can say this quite closes https://github.com/NVIDIA/aistore/issues/273 yet but it's an important step and a good start.

I'll look a bit more at this later, but I gave it a first pass. The main thing I'd like to see is making the config and environment options a bit more generic. That'll make it easier to plug in other similar KVDB services people might want to use. 

External key management is probably a wide enough topic to deserve its own MR, regardless of Redis implementation. Though delegating control to an external key provider does simplify things quite a bit, so likely required at least for now. 

We're a bit hesitant to add new direct dependencies, but I don't know if there's any kvdb abstraction libraries out there that would be as easy to plug in to interface with redis, so it may be acceptable here. I'll need to check on license requirements.

cmd/authn/signing/rsa_mgr.go
aaronnw, line 1: Maybe I missed something here, but if the RSA key is externally managed, does the public JWK make it into the JWKS? If that external key rotates, what triggers the JWKS to update?

shreyaskommuri:
RedisConf → KVServiceConf, field Redis → Service (JSON: database.service.*); env vars renamed to AIS_AUTHN_KV_* — generic for other backends
DB → DBIndex (JSON: db_index)
Extracted defaultKVServiceAddr constant; removed duplicate hardcoded default
AllowKeyGeneration → ExternallyProvisioned in RSAKeyConfig
Added signing.ErrExternallyProvisioned sentinel → rotation now returns 400 instead of 500
Dropped the k8s/doc ref from the startup error message

cmd/authn/config/cmgr.go
```diff
@@ -41,6 +41,10 @@ type (
 	RSAKeyConfig struct {
 		Filepath string
 		Size     int
+		// ExternallyProvisioned signals that the RSA key is managed outside this process
+		// (e.g. mounted from a Kubernetes Secret). Auto-generation and API-driven rotation
+		// are both disabled: a missing key is a configuration error, not a recoverable state.
+		ExternallyProvisioned bool
```
aaronnw: Mentioned in another comment, but I think this is separate and major enough of a new change to warrant its own PR, without the added requirement of dealing with the redis lib and licensing. I'd suggest you just open it separately.

aaronnw (changes requested):

cmd/authn/signing/rsa_mgr.go
shreyaskommuri: Good catch. The JWKS is built from the key on disk during `Init()` and stored in the in-memory bundle — it is only refreshed when `RotateKey()` is called. If an externally managed key file changes (e.g. a Secret is rotated in-place), the process must be restarted to pick up the new key and update the JWKS. Added a comment on `GetJWKS()` to document this. Live reload without a restart would be a follow-up.

aaronnw: I think that's reasonable for now to expect external rotation to require reboot and postpone live reload.

shreyaskommuri:
Addressed all review feedback — resolving threads. Summary of changes made:

**Generic KV config (aaronnw)**
- `RedisConf` → `KVServiceConf`; JSON path is now `auth.db.service.{addr,password,db_index,tls_enabled}`
- Env vars renamed to `AIS_AUTHN_KV_{ADDR,PASSWORD,DB_INDEX}` — usable for any KV backend
- Default address extracted to `defaultKVServiceAddr` package-level constant; no duplicate hardcoded strings

**Redis dependency (aaronnw)**
- BSD-2 license text and copyright added to `NOTICE`

**Externally provisioned RSA key (aaronnw)**
- Split into #294 per request; includes `ExternallyProvisioned` config, fatal startup error on missing key, `ErrExternallyProvisioned` sentinel → rotation returns 400 instead of 500
- K8s doc ref removed from error message
- `GetJWKS()` comment documents that external key rotation requires a process restart; live reload is a follow-up

**Other**
- `DB` field renamed to `DBIndex` (JSON: `db_index`)
- `DatabaseConf.Validate()` sets the default KV service address, removing the need for a duplicate default elsewhere

PR #292, opened 2026-05-01
Update `ais_target_state_flags` Prometheus metric when rebalance or resilver is interrupted

eric-higgins-ai:
I would like to use the `RebalanceInterrupted` bit of the `ais_target_state_flags` Prometheus metric to monitor rebalance stability. However, in the current code it's never set to 1.

Changes:
* The `RebalanceInterrupted` bit in the `ais_target_state_flags` Prometheus metric now gets set when a rebalance is interrupted. It gets cleared when a new rebalance starts.
* Launching a rebalance with the `--cleanup` flag sets the `Rebalancing` bit in the `ais_target_state_flags` Prometheus metric
* Launching a rebalance with the `--cleanup` flag clears the `rebalance-interrupted` state in `ais show cluster`
* Launching a rebalance with the `--cleanup` flag sets the `Rebalancing` bit in Prometheus
* The `ResilverInterrupted` bit in the `ais_target_state_flags` Prometheus metric now gets set when a resilver is interrupted. It gets cleared when a new resilver starts.

### DIFF
--- a/ais/target.go
+++ b/ais/target.go
@@ -428,7 +428,7 @@ func (t *target) Run() error {
 	// When `auth.intra_cluster.enabled`, restarted node must _not_ originate signed traffic to a peer
 	// until that peer has received an (updated version of) Smap containing the node's (current)
 	// verifying key.
-	fatalErr, writeErr := t.checkRestarted(config)
+	fatalErr, writeErr := t.checkStartupMarkers(config)
 	if fatalErr != nil {
 		cos.ExitLog(fatalErr)
 	}
@@ -679,7 +679,7 @@ func rxAnyStream(w http.ResponseWriter, r *http.Request) {
 	transport.RxAnyStream(w, r)
 }
 
-func (t *target) checkRestarted(config *cmn.Config) (fatalErr, writeErr error) {
+func (t *target) checkStartupMarkers(config *cmn.Config) (fatalErr, writeErr error) {
 	if fs.MarkerExists(fname.NodeRestartedMarker) {
 		red := redial{t: t, dialTout: config.Timeout.CplaneOperation.D(), totalTout: config.Timeout.MaxKeepalive.D()}
 		if red.acked() {
@@ -688,6 +688,12 @@ func (t *target) checkRestarted(config *cmn.Config) (fatalErr, writeErr error) {
 		}
 		t.statsT.SetFlag(cos.NodeAlerts, cos.NodeRestarted)
 		fs.PersistMarker(fname.NodeRestartedPrev, false /*quiet*/)
+	}
+	if fs.MarkerExists(fname.RebalanceMarker) {
+		t.statsT.SetFlag(cos.NodeAlerts, cos.RebalanceInterrupted)
+	}
+	if fs.MarkerExists(fname.ResilverMarker) {
+		t.statsT.SetFlag(cos.NodeAlerts, cos.ResilverInterrupted)
 	}
 	fatalErr, writeErr = fs.PersistMarker(fname.NodeRestartedMarker, false /*quiet*/)
 	return
--- a/fs/persistent_md.go
+++ b/fs/persistent_md.go
@@ -118,6 +118,18 @@ func RemoveMarker(marker string, stup cos.StatsUpdater, stopping bool) (ok bool)
 			}
 		}
 	}
+
+	if !ok {
+		switch marker {
+		case fname.RebalanceMarker:
+			stup.SetClrFlag(cos.NodeAlerts, cos.RebalanceInterrupted, cos.Rebalancing)
+		case fname.ResilverMarker:
+			stup.SetClrFlag(cos.NodeAlerts, cos.ResilverInterrupted, cos.Resilvering)
+		}
+
+		return ok
+	}
+
 	switch marker {
 	case fname.RebalanceMarker:
 		stup.ClrFlag(cos.NodeAlerts, cos.RebalanceInterrupted|cos.Rebalancing)
--- a/fs/persistent_md_test.go
+++ b/fs/persistent_md_test.go
@@ -5,8 +5,12 @@
 package fs_test
 
 import (
+	"os"
+	"path/filepath"
+	"sync"
 	"testing"
 
+	"github.com/NVIDIA/aistore/cmn/cos"
 	"github.com/NVIDIA/aistore/cmn/fname"
 	"github.com/NVIDIA/aistore/core/mock"
 	"github.com/NVIDIA/aistore/fs"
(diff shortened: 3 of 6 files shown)


alex-aizman:
I can see where RebalanceInterrupted flag gets set, but I don’t see where it gets cleared...

I guess, the point is to mirror the respective persistent marker (fname.RebalanceMarker).

need to find out how does that work with existing code:

https://github.com/NVIDIA/aistore/blob/main/ais/htrun.go#L216

etc.

and why that's not enough.

eric-higgins-ai:
@alex-aizman the flag gets cleared in [RemoveMarker](https://github.com/NVIDIA/aistore/blob/d8892e1c0cbe4f25eb6133dad57976e12eeddd7f/fs/persistent_md.go#L106), for example [here](https://github.com/NVIDIA/aistore/blob/d8892e1c0cbe4f25eb6133dad57976e12eeddd7f/reb/globrun.go#L256)

I don't have a good enough mental model of the code to know for sure why https://github.com/NVIDIA/aistore/blob/main/ais/htrun.go#L216 isn't sufficient, but my understanding is just that that's never synced to the `ais_target_state_flags` prometheus metric

alex-aizman:
> I would like to use the RebalanceInterrupted bit of the ais_target_state_flags Prometheus metric to monitor rebalance stability. However, in the current code it's never set to 1.

No, that's not true

```
$ git grep RebalanceInterrupted | grep "\bSet"
ais/htcommon.go:580:            nsti.Flags = nsti.Flags.Set(cos.RebalanceInterrupted)
ais/htrun.go:233:                       cm.Flags = cm.Flags.Set(cos.RebalanceInterrupted)
$ 
```

Closing.

eric-higgins-ai:
@alex-aizman those calls don't get propagated to the `ais_target_state_flags` prometheus metric. The prometheus metric is only set via `SetFlag`/`ClrFlag` in `stats/common.go`. If you look for usages of those functions, you'll see that `ClrFlag` is called with `RebalanceInterrupted` [here](https://github.com/NVIDIA/aistore/blob/7ad0be789a6857623e4a69f6c6e76212d344baf0/fs/persistent_md.go#L123) but there's no corresponding `SetFlag` call

eric-higgins-ai:
You can reproduce this by running `ais start rebalance` and then immediately running `ais stop rebalance` to interrupt. If you observe the value of the flag in prometheus with this promql query `floor(ais_target_state_flags / 16) % 2 == 1` then you can see that the flag isn't set when using the code from the `main` branch, and is set when this PR is applied

alex-aizman:
how about `ais show cluster` - much more direct way to see all those alerts:

```
 11 type NodeStateFlags uint64 // same as cos.BitFlags
 12 
 13 const NodeAlerts = "state.flags"
 14 
 15 const noAlerts = "ok"
 16 
 17 const (
 18         VoteInProgress       = NodeStateFlags(1 << iota) // warning
 19         ClusterStarted                                   // info: (primary: cluster-started | all other nodes: joined-cluster)
 20         NodeStarted                                      // info: (started; possibly, not joined yet)
 21         Rebalancing                                      // warning
 22         RebalanceInterrupted                             // warning
 23         Resilvering                                      // warning
 24         ResilverInterrupted                              // warning
 25         NodeRestarted                                    // warning (powercycle, crash)
 26         OOS                                              // node out of space; red alert (see IsRed below)
 27         OOM                                              // out of memory; red alert
 28         MaintenanceMode                                  // warning
 29         LowCapacity                                      // node (used > high); warning: OOS possible soon..
 30         LowMemory                                        // ditto OOM
 31         DiskFault                                        // red
 32         NumGoroutines                                    // yellow
 33         HighNumGoroutines                                // red
 34         CertWillSoonExpire                               // warning X.509
 35         CertificateExpired                               // red --/--
 36         CertificateInvalid                               // red --/--
 37         KeepAliveErrors                                  // warning (new keep-alive errors during the last 5m)
 38         OOCPU                                            // out of CPU; red
 39         LowCPU                                           // warning
 40         DiskOOS                                          // disk out of space
 41         DiskLowCapacity                                  // warning
 42 )
```

does it reproduce with `ais show cluster`?

eric-higgins-ai:
Yeah `ais show cluster` works - I can see the `rebalance-interrupted` flag there. I wanted to access the information in Prometheus though, so we can integrate it with our on-call alerting setup.

alex-aizman:
@gaikwadabhishek take a look

reopening this PR...

reb/globrun.go
```diff
@@ -609,25 +730,16 @@ func (reb *Reb) fini(rargs *rargs, err error, tstats cos.StatsUpdater) {
...
-	// Close and Rx-unregister data mover:
-	// - inbound drain is bounded by transport.Quiescent (Config.Transport.QuiesceTime, default 10s) of inbound silence;
-	// - enforced by UnregRecv inside endStreams;
-	// - endStreams transitions stage to <fin-streams>;
-	// - MUST run lock-free.
-	ok, curStage := reb.endStreams(rargs, err)
-	if !ok {
-		nlog.Warningln(rargs.logHdr, "ended streams when curr. stage:", stages[curStage])
+	} else {
+		tstats.SetFlag(cos.NodeAlerts, cos.RebalanceInterrupted)
```
gaikwadabhishek, line 618: This does not cover cleanup-mode (`--cleanup` in `ais start rebalance`) rebalances. `RunCleanup` creates the same `RebalanceMarker`, but finiCleanup retains it on abort without setting RebalanceInterrupted. I reproduced this by aborting cleanup mode: the marker remained and cluster status reported interruption, while ais_target_state_flags stayed clear. Please update finiCleanup

reb/globrun.go
gaikwadabhishek, line 609: Please handle a false return from `RemoveMarker`. It currently clears RebalanceInterrupted unconditionally even when marker deletion fails. I reproduced ok=false, markerExists=true, and the metric bit cleared. This leaves cluster status and Prometheus inconsistent. The bit should only be cleared after confirmed marker removal, or restored when removal fails

gaikwadabhishek, line 606: Could a stop request arrive after Quiesce() returns but before xreb.Finish()? The marker is already removed here, but the xaction is still running, so the stop is accepted. Finish() then records it as aborted while the marker and Prometheus flag indicate success. Can we either reject late stops or update the marker after the final xaction state is known?

gaikwadabhishek (changes requested):
Hey @eric-higgins-ai , thanks for contributing. Few edge scenarios are not handled or needs exploration. In re: prometheus, the alert already exists - `AISRebalanceInterrupted` @ ais-k8s/monitoring/kube-prom/alert-rules/templates/native-rules.yaml. 

I think if you have time this change can also be applied to `resilver interrupted` scenario.

reb/globrun.go
gaikwadabhishek: this scenario is pre-existing, technically not in your scope as well. so skip this for now. but needs fixing so add a TODO atleast

gaikwadabhishek: the comment is for the below line

reb/globrun.go
gaikwadabhishek: @eric-higgins-ai you marked this as resolved but no changes were implemented? any issue?

eric-higgins-ai: ah sorry, I resolved the comments for my own bookkeeping but I haven't pushed changes yet as I'm still testing them. I can unresolve the comments

eric-higgins-ai:
ok @gaikwadabhishek I addressed your comments. I also found a few other bugs while testing, which I also fixed:
* When a new rebalance starts, the `rebalance-interrupted` flag gets cleared in the output of `ais show cluster` but doesn't get cleared in Prometheus until the rebalance finishes successfully
* When launching a rebalance with `ais start rebalance --cleanup`, there were 2 issues:
  * The `rebalance-interrupted` flag doesn't get cleared in `ais show cluster` or in prometheus. This is inconsistent with the behavior without the "cleanup" flag
  * The `rebalancing` flag doesn't get set in prometheus

gaikwadabhishek:
Thanks @eric-higgins-ai. Changes looks good. Let me verify locally whether everything is working right (want to try some edge cases) and then I will approve.

fs/persistent_md.go
```diff
@@ -118,6 +118,11 @@ func RemoveMarker(marker string, stup cos.StatsUpdater, stopping bool) (ok bool)
 			}
 		}
 	}
+
+	if !ok {
+		return ok
+	}
```
gaikwadabhishek, line 131: (extreme edge case) when marker deletion fails, RemoveMarker returns without restoring the interrupted flag.

A new cleanup/rebalance/resilver already cleared the previous interrupted flag at startup.

 I reproduced this on a live three-target cluster:
      - Before cleanup: flags 150, including RebalanceInterrupted.
      - Started cleanup: the stale interrupted flag was cleared as intended.
      - Forced marker deletion to fail. (just removed permission to not be able to write to that dir)
      - Cleanup completed with flags 134, missing RebalanceInterrupted.
      - The marker remained, and ais show cluster still reported 
[rebalance-interrupted restarted].

Prometheus and persistent cluster state remain inconsistent.

gaikwadabhishek: minimal fix should remain centralized in RemoveMarker: on failure, set the appropriate interrupted bit and clear the corresponding running bit.

gaikwadabhishek: add/change tests (if possible) not exactly this test as its very very edge case. see fs/persistent_md_test.go (add failure path assertions)

res/resilver.go
```diff
@@ -239,6 +245,7 @@ func wait(jgroup *mpather.Jgroup, xres *xs.Resilver, tstats cos.StatsUpdater) {
 		select {
 		case <-xres.ChanAbort():
 			_ = jgroup.Stop()
```
gaikwadabhishek: can choose the finished branch when abort and completion become ready together.. Same TODO I think (pls verify once whether that race thing applies here as well)

gaikwadabhishek (changes requested):
cleanup-mode interruption and resilver looks good. marker deletion still needs work

gaikwadabhishek:
PR is 95% complete, almost there 👍

res/resilver.go
eric-higgins-ai: I have no idea how to reproduce the case, so I read through the code with the help of AI and it does seem like the race condition exists. I'll add a comment

fs/persistent_md.go
eric-higgins-ai: I think I addressed this, but I might have misunderstood your desired solution. Let me know if I got the right idea

fs/persistent_md_test.go
```diff
@@ -69,3 +73,132 @@ func TestMarkers(t *testing.T) {
...
+)
+
+var _ cos.StatsUpdater = (*flagTrackingStats)(nil)
+
+func (s *flagTrackingStats) Get(string) int64                  { return 0 }
+func (s *flagTrackingStats) Inc(string)                        {}
+func (s *flagTrackingStats) IncWith(string, map[string]string) {}
+func (s *flagTrackingStats) Add(string, int64)                 {}
+func (s *flagTrackingStats) Observe(string, float64)           {}
+func (s *flagTrackingStats) AddWith(...cos.NamedVal64)         {}
```
gaikwadabhishek: remove the unused receiver names from lines 104–109

gaikwadabhishek (changes requested):
looks good. minor lint failures

fs/persistent_md_test.go
eric-higgins-ai: :facepalm: sorry, I forgot to run lint. Should be fixed now

gaikwadabhishek:
nit: update commit message... mention rebalance resilver both.. add a small desc in bullets saying what all was changed. should be concise. check the other commits of this repo for the pattern

fs/persistent_md_test.go
gaikwadabhishek: its still there, not sure if you pushed it yet. hasnt reflected

gaikwadabhishek (approved):

fs/persistent_md_test.go
eric-higgins-ai: We do need these `Get`, `Inc`, etc. functions to implement the interface, so it was just the receiver names that were causing the lint errors. I removed those and lint passed when I ran it locally

eric-higgins-ai:
@gaikwadabhishek updated the PR description and commit message, let me know if anything looks wrong

ais/target.go
```diff
@@ -670,15 +681,6 @@ func (t *target) initRecvHandlers() {
...
-func rxAnyStream(w http.ResponseWriter, r *http.Request) {
-	if !reqIsIntra(r) {
-		err := fmt.Errorf("invalid request over %s: %s %s from %s", reqNetName(_reqNet(r)), r.Method, r.URL.Path, r.RemoteAddr)
-		cmn.WriteErr(w, r, err, http.StatusForbidden)
-		return
-	}
-	transport.RxAnyStream(w, r)
-}
-
 func (t *target) checkRestarted(config *cmn.Config) (fatalErr, writeErr error) {
```
alex-aizman: nit: rename it as `checkStartupMarkers`

(or pick a better name - but it's not only about "restarted" anymore)

eric-higgins-ai: updated!

fs/persistent_md.go
```diff
@@ -118,6 +118,18 @@ func RemoveMarker(marker string, stup cos.StatsUpdater, stopping bool) (ok bool)
...
 		}
 	}
+
+	if !ok {
+		switch marker {
+		case fname.RebalanceMarker:
+			stup.SetClrFlag(cos.NodeAlerts, cos.RebalanceInterrupted, cos.Rebalancing)
+		case fname.ResilverMarker:
+			stup.SetClrFlag(cos.NodeAlerts, cos.ResilverInterrupted, cos.Resilvering)
+		}
```
alex-aizman, line 128: so, we have just accomplished all the data-moving work, and now we just failed to remove a measly little zero-size file... 

this is defensible if you say: persistent state must be in total sync with Prometheus

but I still have to ask... because it still blows my mind

eric-higgins-ai: As a user of AIStore I would expect the prometheus metrics to always be in sync with whatever is shown in `ais show cluster`.

This edge case is ultimately small though, so I'm ok with whatever behavior y'all prefer here

alex-aizman:
Eric,

Thank you!

PR #293, opened 2026-05-06
python/etl: stream HTTPMultiThreadedServer no-FQN hpush PUT

chanu1406:
Follow up to #289 — stream no-FQN `hpush` PUT requests in `HTTPMultiThreadedServer` in constant memory.

Previously `_get_stream_reader` buffered the full request body into a `BytesIO` before `transform_stream` ran. For large objects this is the same OOM-at-scale concern the GET-side fix addressed.

The HTTP server case is simpler than the async FastAPI side: `self.rfile` is already a sync `BinaryIO`, so there's no async/sync bridge needed. The fix is a thin length-bounded wrapper (`_RFileLimitedReader`) that gives `transform_stream` the correct EOF semantics without the full-body allocation.

**Changes:**
- `_RFileLimitedReader`: sync `io.RawIOBase` subclass that clamps `read()` to `Content-Length` remaining bytes
- `_direct_put_stream_with_retry`: gate retries on replayability (`bool(fqn) or is_get`); no-FQN PUT bodies are one-shot so local retries are skipped and a transient direct-put error surfaces to AIS as a transform failure
- Tests: 4 new reader-behavior tests, 3 new retry-gate tests, 3 existing tests converted to the replayable GET path, 3 BytesIO-seek-replay tests retired

### DIFF
--- a/python/aistore/sdk/etl/webserver/http_multi_threaded_server.py
+++ b/python/aistore/sdk/etl/webserver/http_multi_threaded_server.py
@@ -2,11 +2,11 @@
 # Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
 #
 
+import io
 import time
 from http.server import HTTPServer, BaseHTTPRequestHandler
-from io import BytesIO
 from socketserver import ThreadingMixIn
-from typing import Iterator, Type, Tuple
+from typing import BinaryIO, Iterator, Type, Tuple
 import signal
 import threading
 from urllib.parse import urlparse, parse_qs
@@ -20,6 +20,7 @@ from aistore.sdk.etl.webserver.base_etl_server import (
     RETRY_BACKOFF_BASE,
     RETRY_BACKOFF_MAX,
     _handle_direct_put_transient_error,
+    _compute_replayable_retries,
 )
 from aistore.sdk.etl.webserver.utils import (
     compose_etl_direct_put_url,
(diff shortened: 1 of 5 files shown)


python/aistore/sdk/etl/webserver/http_multi_threaded_server.py
```diff
@@ -40,6 +40,43 @@
...
+    end of this request body. Passing it directly to `transform_stream` would
+    cause any transform that calls `reader.read()` with no size argument to
+    block indefinitely waiting for the client to close the connection.
+
+    This wrapper tracks `Content-Length` remaining bytes and clamps every
+    `read()` call accordingly, giving transforms the same EOF semantics they
+    get from a `BytesIO` — without buffering the full body upfront.
+
+    The request body is one-shot; `_direct_put_stream_with_retry` sets
+    `effective_retries=0` on this path. Transforms are expected to consume
```
gaikwadabhishek: I am working on a MR that will introduce a new status code 503 + a header which will tell AIS to retry the entire PUT operation again. I think we will have to use that here, I will ping that to you once its ready

python/aistore/sdk/etl/webserver/http_multi_threaded_server.py
```diff
@@ -208,15 +244,17 @@ def _direct_put_stream_with_retry(  # pylint: disable=too-many-arguments,too-man
...
-            _get_stream_reader and seeked back to 0 on retry so the same bytes are
-            replayed (mirrors FastAPIServer._direct_put_stream_with_retry).
+            Replayable sources (FQN-backed or GET) close and reopen the reader on
+            each retry. No-FQN PUT bodies are one-shot (request body is consumed
+            from the socket); effective_retries is forced to 0 and a transient
+            direct-put error surfaces to AIS as a transform failure.
             """
             etl = self.server.etl_server
+            replayable = bool(fqn) or is_get
+            effective_retries = etl.direct_put_retries if replayable else 0
```
gaikwadabhishek: maybe lets create a util for this and use it in both the files fastapi and this

python/aistore/sdk/etl/webserver/http_multi_threaded_server.py
```diff
@@ -225,22 +263,25 @@ def _direct_put_stream_with_retry(  # pylint: disable=too-many-arguments,too-man
                             raw_path,
                         )
                     except ETLDirectPutTransientError as exc:
-                        if attempt >= etl.direct_put_retries:
+                        if attempt >= effective_retries:
+                            if not replayable and etl.direct_put_retries:
```
gaikwadabhishek, line 283: over here you will have to return the other http code and set the return header. I will ping you the commit for ref once its merged

python/aistore/sdk/etl/webserver/http_multi_threaded_server.py
```diff
@@ -40,6 +40,43 @@
...
+            return b""
+        if size is None or size < 0:
+            data = self._rfile.read(self._remaining)
+            self._remaining = 0
+            return data
+        to_read = min(size, self._remaining)
+        data = self._rfile.read(to_read)
+        self._remaining -= len(data)
+        return data
+
```
gaikwadabhishek, line 80: `_RFileLimitedReader.close()` is inherited from io.RawIOBase — it does not drain `_remaining` from `self._rfile`. If `transform_stream` returns early (e.g., transform reads a header, decides to fail, raises), the leftover request body bytes are still sitting on the socket. On a keep-alive connection the next request will fail to parse because the HTTP parser sees mid-body bytes as the start of a request line. 

The class docstring documents "Transforms are expected to consume the full body" but doesn't enforce it. The BytesIO-based code didn't have this hazard because the body was always fully read upfront. 

This is a real behavior change, not a wash. Recommend overriding `close()` to drain remaining bytes 

smth like

  def close(self) -> None:
      while self._remaining > 0:
          chunk = self._rfile.read(min(self._remaining, 65536))
          if not chunk:
              break
          self._remaining -= len(chunk)
      super().close()

gaikwadabhishek (changes requested):
LGTM. just fix the close. If my commit lands before yours, then I expect you to add the hdr and change the status code of the put failure to retry. otherwise we can do it in another PR

chanu1406:
@gaikwadabhishek, thanks for the review. I will address the close and implement the shared util.

gaikwadabhishek (approved):
PR looks complete, I will ping you once I merge the other change about hdr. Can you resolve other comments. Ok to merge

chanu1406:
> PR looks complete, I will ping you once I merge the other change about hdr. Can you resolve other comments. Ok to merge

Resolved @gaikwadabhishek

PR #294, opened 2026-05-06
authn: externally provisioned RSA key config

shreyaskommuri:
## Summary

Adds `auth.externally_provisioned` (config file) / `AIS_AUTHN_EXTERNALLY_PROVISIONED` (env var) to signal that the RSA signing key is managed outside the AuthN process.

When set:
- A missing key file at startup is a **fatal configuration error**, not a recoverable state (no auto-generation)
- `POST /v1/auth/keys/rotate` returns **400 Bad Request** via the `signing.ErrExternallyProvisioned` sentinel rather than 500

This is useful for any deployment requiring a shared, pre-provisioned key — multi-replica setups, secret managers, HSMs, etc. — and is **independent of the storage backend** in use.

Split out from #290 per reviewer request.

## Test plan
- [ ] Start AuthN with `externally_provisioned: true` and a valid key file — should load normally
- [ ] Start AuthN with `externally_provisioned: true` and no key file — should exit with a clear error
- [ ] Call `POST /v1/auth/keys/rotate` with `externally_provisioned: true` — should return 400
- [ ] Start AuthN with `AIS_AUTHN_EXTERNALLY_PROVISIONED=true` env var — env should override config file value

### DIFF
--- a/api/authn/config.go
+++ b/api/authn/config.go
@@ -40,6 +40,14 @@ const (
 	defaultPort             = 52001
 )
 
+// RSA key management modes for auth.rsa.mode
+const (
+	// RSAModeExternal signals that the RSA key is managed outside this process
+	// (e.g. mounted from a secret manager or HSM). Auto-generation and API-driven
+	// rotation are disabled; a missing key file is a fatal configuration error.
+	RSAModeExternal = "external"
+)
+
 type (
 	Config struct {
 		Server  ServerConf  `json:"auth"`
@@ -70,10 +78,14 @@ type (
 		// Also used to determine max-age for client caches of JWKS
 		Expire cos.Duration `json:"expiration_time"`
 		// Only used for validating RSA public key against AIS clusters
-		PubKey *string `json:"public_key"`
-		// Size of RSA private key to generate
-		RSAKeyBits int          `json:"rsa_key_bits"`
-		DBConf     DatabaseConf `json:"db"`
+		PubKey *string      `json:"public_key"`
+		RSA    RSAConf      `json:"rsa"`
+		DBConf DatabaseConf `json:"db"`
+	}
+	RSAConf struct {
+		Bits int `json:"bits,omitempty"`
+		// "external": key is managed outside this process; auto-generation and API rotation are disabled
+		Mode string `json:"mode,omitempty"`
 	}
 	DatabaseConf struct {
 		DBType   string `json:"type"`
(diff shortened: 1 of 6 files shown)


alex-aizman:
I'd maybe look for something more structured, like e.g.:

```
"auth": {
  "expiration_time": "24h",
  "rsa": {
    "bits": 2048,
    "mode": "external"
  },
  "db": {
    "type": "...",
    "filepath": "..."
  }
}
```

This would be cleaner than a yet another boolean knob at the top level:

```
"externally_provisioned": true
```

alex-aizman:
Secondly and separately, your commit is called "externally provisioned" but then - what about external rotation? There's nothing in the commit itself - unless I overlooked

alex-aizman:
Finally - yes, the motivation is clear and makes sense.

aaronnw:
> Secondly and separately, your commit is called "externally provisioned" but then - what about external rotation? There's nothing in the commit itself - unless I overlooked

related comment from the initial MR https://github.com/NVIDIA/aistore/pull/290#discussion_r3165903393 
should support eventually, but I agree it's not a hard requirement for the initial implementation.

shreyaskommuri:
Thanks for the feedback — addressed both points.

**Structural refactor (alex-aizman)**

Replaced the top-level `externally_provisioned bool` with a nested `rsa` block under `auth`, as suggested:

```json
"auth": {
  "expiration_time": "24h",
  "rsa": {
    "bits": 2048,
    "mode": "external"
  },
  "db": {
    "type": "...",
    "filepath": "..."
  }
}
```

`rsa_key_bits` moves to `rsa.bits`; `externally_provisioned: true` becomes `rsa.mode: "external"`. The env var changes from `AIS_AUTHN_EXTERNALLY_PROVISIONED=true` (bool) to `AIS_AUTHN_RSA_MODE=external` (string). `Validate()` rejects any unrecognized mode value. The internal `RSAKeyConfig.ExternallyProvisioned` bool is unchanged — it is derived from the mode string in `GetRSAConfig()`.

**External rotation (alex-aizman / aaronnw)**

The JWKS is built from the key file during `Init()` and held in-memory; it is not reloaded at runtime. To pick up an externally rotated key, replace the file on disk and restart the process. Documented this directly on `Init()` so the constraint is visible at the call site. Live reload without a restart is a follow-up, per the discussion in #290.

api/authn/config.go
```diff
@@ -70,10 +78,14 @@ type (
 		// Also used to determine max-age for client caches of JWKS
 		Expire cos.Duration `json:"expiration_time"`
 		// Only used for validating RSA public key against AIS clusters
-		PubKey *string `json:"public_key"`
-		// Size of RSA private key to generate
-		RSAKeyBits int          `json:"rsa_key_bits"`
-		DBConf     DatabaseConf `json:"db"`
+		PubKey *string      `json:"public_key"`
+		RSA    RSAConf      `json:"rsa"`
```
aaronnw, line 82: Just noting this is breaking, but since it's a recent addition and authN is still marked as under development, I think it's worth keeping now.

aaronnw (approved):

aaronnw (changes requested):
Please squash to one commit and LGTM!

PR #299, opened 2026-05-22
startup: allow primary to join after storage loss

### LINKED ISSUE #298: Original primary proxy can't rejoin the cluster after losing state storage

- [x] I have searched the existing issues

### Describe the bug

If the primary proxy dies + loses its metadata storage, the cluster will elect a new primary. However, when the original primary tries to rejoin the cluster, it will see in the global configmap that it should be primary, and will try to start a new cluster. This fails with this error:
```
E 14:58:53.356972 earlystart:334 FATAL ERROR: p[D00x2v6n8mdwd]cannot create cluster with no targets,Smap v1[, p[D00x2v6n8mdwd], t=0, p=1] 
```

### Expected Behavior

The original primary should either just join as a secondary, or it should load the metadata from the existing cluster and then request to become primary.

### Current Behavior

The original primary isn't able to rejoin the cluster.

### Steps To Reproduce

I'm using the k8s operator so these steps assume that as well.

1. Create a cluster with multiple proxies
2. Delete the primary proxy's PVC
3. Delete the primary proxy pod
4. Observe that it continually restarts and fails to join the cluster

### Possible Solution

_No response_

### Additional Information/Context

I'm happy to open a PR to fix this, I've just been looking at it for a while and I'm not confident I know the correct approach so I want to get some input.

### AIStore build/version

v4.5

Oracle Kubernetes Engine 1.34.1, Oracle Linux 8.10

eric-higgins-ai:
This fixes https://github.com/NVIDIA/aistore/issues/298 by checking for an existing cluster before joining as primary. If one exists then we join as secondary, and if one doesn't exist then we start a new one.

### DIFF
--- a/ais/earlystart.go
+++ b/ais/earlystart.go
@@ -6,9 +6,12 @@ package ais
 
 import (
 	"fmt"
+	"net/http"
 	"net/url"
 	"runtime"
 	"time"
+
+	jsoniter "github.com/json-iterator/go"
 
 	"github.com/NVIDIA/aistore/api/apc"
 	"github.com/NVIDIA/aistore/api/env"
@@ -146,6 +149,16 @@ func (p *proxy) determineRole(smap *smapX /*loaded*/, config *cmn.Config) (prim 
 			config.Proxy.PrimaryURL == p.si.URL(cmn.NetPublic)
 		if !prim.isCfg {
 			prim.isCfg = p.si.HasURL(config.Proxy.PrimaryURL)
+		}
+		// if config designates this node as primary but there is no local Smap,
+		// probe well-known URLs to detect an already-running cluster; if found
+		// with a different primary, join as secondary rather than attempting
+		// to create a new cluster that would conflict with the existing one.
+		if prim.isCfg {
+			if existingPrimary := p.probeForCluster(config); existingPrimary != "" {
+				prim.isCfg = false
+				prim.url = existingPrimary
+			}
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


ais/earlystart.go
```diff
@@ -147,6 +150,16 @@ func (p *proxy) determineRole(smap *smapX /*loaded*/, config *cmn.Config) (prim
 		if !prim.isCfg {
 			prim.isCfg = p.si.HasURL(config.Proxy.PrimaryURL)
 		}
+		// if config designates this node as primary but there is no local Smap,
+		// probe well-known URLs to detect an already-running cluster; if found
+		// with a different primary, join as secondary rather than attempting
+		// to create a new cluster that would conflict with the existing one.
+		if prim.isCfg {
+			if existingPrimary := p.probeForCluster(config); existingPrimary != "" {
```
alex-aizman, line 158: the big assumption here is whether we can trust this config, and whether there may exist a higher-version config in the cluster that'd say otherwise

OTOH, this is a well-known (and quite old) problem: starting-up as a primary without a cluster map carries this inevitable risk of creating an empty cluster, or worse - a full-blown split-brain.

I'll think about it..

ais/earlystart.go
```diff
@@ -147,6 +150,16 @@ func (p *proxy) determineRole(smap *smapX /*loaded*/, config *cmn.Config) (prim
...
+				prim.isCfg = false
+				prim.url = existingPrimary
```
alex-aizman, line 160: separately, just one technical question: would it make sense to patch it _after_ you return from `determineRole` with a `nil` Smap?

more precisely, the condition I'd check upon return would be something like:

```
smap == nil && (prim.isCfg || prim.isEP)
```

ais/earlystart.go
```diff
@@ -1021,6 +1034,71 @@ ret:
...
 
+// probeForCluster probes the configured discovery URL to detect
+// an already-running AIStore cluster.
+func (p *proxy) probeForCluster(config *cmn.Config) (primaryURL string) {
+	var (
+		selfPub  = p.si.URL(cmn.NetPublic)
+		selfCtrl = p.si.URL(cmn.NetIntraControl)
+	)
+
+	nsti, err := p._probeOne(config.Proxy.DiscoveryURL)
```
alex-aizman, line 1045: I wouldn't single out this one - there's the entire config section that's intended exactly for this type scenarios:

```
$ ais config cluster proxy --json
{
    "proxy": {
        "original_url": ...,
        "discovery_url": ...,
        "non_electable": false,
        "primary_url": ...
    }
} 
```

ais/earlystart.go
```diff
@@ -1021,6 +1034,71 @@ ret:
...
+		Query:  url.Values{apc.QparamClusterInfo: []string{"true"}},
+	}
+	cargs.timeout = cmn.Rom.CplaneOperation()
+	// p.owner.smap.get() may return nil before bootstrap puts anything in;
+	// pass a blank smapX so that setIntraHdrs does not dereference a nil pointer.
+	smap := p.owner.smap.get()
+	if smap == nil {
+		smap = newSmap()
+	}
+	res := p.call(cargs, smap)
```
alex-aizman, line 1090: I didn't get this bit... we are here because there's no Smap in the first place, and so I'd expect to be making a call with an empty instance...

> pass a blank smapX

exactly. Something like:

```
var empty smapX
```

ais/earlystart.go
```diff
@@ -1021,6 +1034,71 @@ ret:
...
+		selfCtrl = p.si.URL(cmn.NetIntraControl)
+	)
+
+	nsti, err := p._probeOne(config.Proxy.DiscoveryURL)
+	if err != nil {
+		nlog.Infof("%s: probe %s: %v", p, config.Proxy.DiscoveryURL, err)
+		return ""
+	}
+	if nsti.Smap.UUID == "" || nsti.Smap.Version == 0 || nsti.Smap.Primary.ID == "" {
+		return ""
```
alex-aizman, line 1051: there are two related helpers - pick one:

```
ais/clustermap.go:242:func (m *smapX) isValid() bool {
ais/clustermap.go:259:func (m *smapX) validate() error {
```

alex-aizman:
separately:

1. I wonder how would you test it?
2. this is non-trivial complexity,  I'm impressed

ais/earlystart.go
alex-aizman: secondly, make sure that `nsti.Smap.UUID == config.UUID`

this is actually important - not to be making existential decisions based on foreign cluster's metadata

a quick grep will tell the story:

```
$ git grep config.UUID
ais/earlystart.go:633:                          cos.ExitLogf("%s: invalid config UUID: %s", p, cm.Config)
ais/earlystart.go:635:                  if cos.IsValidUUID(config.UUID) && config.UUID != cm.Config.UUID {
ais/gconfig.go:54:func (config *globalConfig) uuid() string   { return config.UUID }
ais/htrun.go:1768:              logmsync(config.Version, newConfig, msg, sender, newConfig.String(), config.UUID)
ais/htrun.go:2021:      if config.UUID != "" && config.UUID != newConfig.UUID {
ais/htrun.go:2022:              err = fmt.Errorf("%s: cluster configs have different UUIDs: (curr %q vs new %q)", ciError(110), config.UUID, newConfig.UUID)
ais/tgtcp.go:1059:              debug.Assert(config.Version > 0 && config.UUID == smap.UUID, config.String(), " vs ", smap.StringEx())
cmn/config.go:2945:             debug.Assert(config.Version > 0 && config.UUID != "")
$ 
```

ais/earlystart.go
```diff
@@ -1021,6 +1034,71 @@ ret:
...
+	}
+	// if the existing cluster still considers us primary, let primaryStartup
+	// handle recovery (other nodes will re-register with us)
+	if nsti.Smap.Primary.ID == p.SID() || pURL == selfPub || pURL == selfCtrl {
+		nlog.Infof("%s: probed %s - existing cluster (UUID=%s v%d) still has self as primary",
+			p, config.Proxy.DiscoveryURL, nsti.Smap.UUID, nsti.Smap.Version)
+		return ""
+	}
+	nlog.Warningf("%s: detected existing cluster (UUID=%s v%d, primary=%s) via %s - joining as secondary",
+		p, nsti.Smap.UUID, nsti.Smap.Version, nsti.Smap.Primary.ID, config.Proxy.DiscoveryURL)
```
alex-aizman, line 1066: what if remote config's version is lower than yours?

assuming of course the same cluster (see my comment above)

alex-aizman:
@eric-higgins-ai I started working on it now.. must be able to finish tomorrow

alex-aizman:
done by:
* 694b34b858bd1b5536
* 5c7e048924f0812f643

eric-higgins-ai:
thanks for your help!

alex-aizman:
separately, I maybe missed to mention related `ais cluster set-primary --force` - the command that allows to forcefully join two clusters (at your own risk, of course)

PR #300, opened 2026-05-27
fix: rename NewErrInvalidaMountpath -> NewErrInvalidMountpath (typo)

immanuwell:
`NewErrInvalidaMountpath` has a stray `a` that snuck in between `Invalid` and `Mountpath`. The error type itself is `ErrInvalidMountpath` (correct), but the constructor got the extra char. Bit embarrassing tbh.

**Files changed:** `cmn/err.go` (definition) + `cmn/config.go` (3 call sites). All callers are within the `cmn` package so no breakage outside.

**To reproduce:**
```go
// grep for the old name - it compiles fine but looks wrong
grep NewErrInvalidaMountpath cmn/err.go
// => func NewErrInvalidaMountpath(mpath, cause string) *ErrInvalidMountpath
```

The struct is `ErrInvalidMountpath`, the constructor should match.

**Fix:** rename the func + update 3 call sites in `ValidateMpath`. Zero behavior change, just fixing the name.

### DIFF
--- a/cmn/config.go
+++ b/cmn/config.go
@@ -2449,13 +2449,13 @@ func ValidateMpath(mpath string) (string, error) {
 	cleanMpath := filepath.Clean(mpath)
 
 	if cleanMpath[0] != filepath.Separator {
-		return mpath, NewErrInvalidaMountpath(mpath, "mountpath must be an absolute path")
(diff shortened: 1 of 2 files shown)


alex-aizman:
yes, a typo..

gaikwadabhishek (approved):
LGTM

PR #303, opened 2026-06-05
etl: forward etl_args across pipeline stages

chanu1406:
This fixes `etl_args` propagation across direct-put ETL pipeline hops.

Before this change, the first ETL stage received per-request transform args from the target request, but downstream stages only received the transformed bytes and remaining pipeline header. As a result, stages 2..N saw empty `etl_args`.

Changes:
- Forward incoming `etl_args` through Go ETL direct-put and WebSocket pipeline paths.
- Forward incoming `etl_args` through Python FastAPI, Flask, and `HTTPMultiThreadedServer` direct-put paths.
- Preserve existing direct-put destination query params such as `uuid` / `owt`.
- Add Go and Python unit coverage for direct-put argument forwarding.

### DIFF
--- a/ext/etl/webserver/webserver.go
+++ b/ext/etl/webserver/webserver.go
@@ -169,7 +169,8 @@ func (base *etlServerBase) getHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 func (base *etlServerBase) handleRequest(objReader io.ReadCloser, w http.ResponseWriter, r *http.Request) {
-	transformedReader, size, err := base.Transform(objReader, r.URL.Path, r.URL.Query().Get(apc.QparamETLTransformArgs))
+	etlArgs := r.URL.Query().Get(apc.QparamETLTransformArgs)
+	transformedReader, size, err := base.Transform(objReader, r.URL.Path, etlArgs)
 	if err != nil {
 		cmn.WriteErr(w, r, fmt.Errorf("[%s] %w", "failed to transform", err))
 		return
@@ -177,7 +178,7 @@ func (base *etlServerBase) handleRequest(objReader io.ReadCloser, w http.Respons
 	objReader.Close() // objReader is supposed to be fully consumed by the custom transform function - safe to close at this point
 
 	if directPutURL := r.Header.Get(apc.HdrNodeURL); directPutURL != "" {
-		base.handleDirectPut(w, transformedReader, size, r.URL.Path, directPutURL)
+		base.handleDirectPut(w, transformedReader, size, r.URL.Path, directPutURL, etlArgs)
 		return
 	}
 
@@ -251,10 +252,10 @@ func (base *etlServerBase) handleWebsocketMessage(conn *websocket.Conn, ctrl etl
 	}
 	defer reader.Close()
 
-	return base.handleWebsocketPipeline(conn, transformed, size, ctrl.Path, ctrl.Pipeline)
+	return base.handleWebsocketPipeline(conn, transformed, size, ctrl.Path, ctrl.Pipeline, ctrl.Targs)
 }
 
-func (base *etlServerBase) handleWebsocketPipeline(conn *websocket.Conn, transformed io.ReadCloser, size int64, objPath, pipeline string) (err error) {
+func (base *etlServerBase) handleWebsocketPipeline(conn *websocket.Conn, transformed io.ReadCloser, size int64, objPath, pipeline, etlArgs string) (err error) {
 	// No pipeline, send transformed data
 	if pipeline == "" {
 		writer, _ := conn.NextWriter(websocket.BinaryMessage)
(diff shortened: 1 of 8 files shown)


python/aistore/sdk/etl/webserver/utils.py
```diff
@@ -86,11 +89,20 @@ def compose_etl_direct_put_url(
         # Case 1: pipeline stage → append object path
         final_path = obj_path
 
+    # Keep xid/stats query params from the destination, then forward etl_args so
+    # downstream stages receive the same per-request args (only stage 1 gets them
+    # from the target otherwise). The final destination-target hop ignores it.
+    query = direct.query
+    if etl_args:
+        query_pairs = parse_qsl(direct.query, keep_blank_values=True)
+        query_pairs.append((QPARAM_ETL_ARGS, etl_args))
```
Nahemah1022, line 102: `compose_etl_direct_put_url` currently appends `etl_args` to the existing query string. In the normal AIS-generated pipeline path this likely won’t duplicate anything, but it would be cleaner and safer to treat `etl_args` as a single-valued param and replace any existing value rather than produce `...?etl_args=old&etl_args=new`.

Please switch this to set/replace semantics, matching the Go implementation, while still preserving unrelated query params like uuid/owt. A small unit test with an input URL that already contains `etl_args` would cover this

python/tests/unit/sdk/test_etl_webserver.py
```diff
@@ -3151,3 +3153,137 @@ def test_bare_connection_error_still_retried(self):
...
+
+        with patch("requests.Session.put") as mock_put:
+            mock_put.return_value = MagicMock(status_code=200, content=b"")
+            response = client.put(
+                "/test/object?etl_args=jpeg", data=b"input data", headers=headers
+            )
+
+            self.assertEqual(response.status_code, 204)
+            called_url = mock_put.call_args.args[0]
+            self.assertIn("etl_args=jpeg", called_url)
```
Nahemah1022: the assertion is a bit too loose: it would pass even if the URL contains duplicate params like `etl_args=old&etl_args=jpeg`. maybe parse the URL and assert the exact query value.

python/aistore/sdk/etl/webserver/fastapi_server.py
gaikwadabhishek: NIT: can you add the .get(..., "").strip() here as well. Just noticed this does it differently

Nahemah1022:
@chanu1406 Excellent PR. This is a genuinely useful enhancement for ETL pipelines, and the implementation is very clean overall.

I only left a couple of minor nits around query-param handling/test strictness. Please address those, and this should be good to merge. Nice work!

ext/etl/webserver/webserver.go
```diff
@@ -399,7 +400,16 @@ func (base *etlServerBase) directPut(directPutURL string, r io.ReadCloser, size
...
+	// Carry etl_args to the next pipeline stage; the receiving ETL server reads it
+	// from the incoming query (see handleRequest). Merge so the destination target's
+	// xid/owt query params (final hop) are preserved.
+	if etlArgs != "" {
+		q := direct.Query()
+		q.Set(apc.QparamETLTransformArgs, etlArgs)
+		finalURL.RawQuery = q.Encode()
+	} else {
+		finalURL.RawQuery = direct.RawQuery
+	}
```
gaikwadabhishek, line 414: nit: but why we need to encode?

python/CHANGELOG.md
```diff
@@ -25,12 +22,12 @@ We structure this changelog in accordance with [Keep a Changelog](https://keepac
...
-- `Job.wait()` resolves the job kind from the cluster when only a job id is
-  given (raising `JobInfoNotFound` if the id is unknown), so a job created
-  with just an id still dispatches correctly.
-- `Job.wait()` / `wait_for_idle()` require an idle-kind job to report idle on
-  2 consecutive polls before completing (mirrors Go's
-  `xact.numConsecutiveIdle`); an abort on any target still returns immediately.
+
+### Fixed
+
+- ETL webservers (FastAPI, Flask, `HTTPMultiThreadedServer`) now forward
```
gaikwadabhishek: nit: why is only HTTPMultiThreadedServer in ticks ``? drop the () 
```suggestion
- ETL webservers now forward ...
```

gaikwadabhishek (changes requested):
LGTM

ext/etl/webserver/webserver.go
chanu1406: etlArgs arrives URL-decoded and may contain characters like {} that need to be re-encoded when written back to RawQuery

chanu1406:
@Nahemah1022 @gaikwadabhishek thanks for the review, I've pushed a commit to address the feedback

gaikwadabhishek (approved):
LGTM

gaikwadabhishek:
will run this pipeline internally once and then merge

gaikwadabhishek:
@chanu1406 can you stash your commits into one commit? and add a signed off by?

chanu1406:
> @chanu1406 can you stash your commits into one commit? and add a signed off by?

Done

PR #304, opened 2026-06-07
authn: test externally provisioned RSA key behavior

Rohithmatham12:
## Summary
- Add coverage for externally provisioned RSA mode failing fast when the configured key file is missing.
- Add coverage that API-driven key rotation is rejected with signing.ErrExternallyProvisioned while preserving the loaded key bundle.
- Exercise the behavior added in #294, which is one of the prerequisites for multi-replica AuthN deployments in #273.

## Testing
- gofmt cmd/authn/signing/rsa_mgr_test.go
- go test ./cmd/authn/signing


aaronnw:
Thanks for the fix! Merged via our internal pipeline, closing PR.

PR #306, opened 2026-06-08
Makefile: fix a bug where GOPATH might be undefined

findlayfeng:
If GOPATH is not defined, the build will use the
incorrect default BUILD_DEST.


alex-aizman:
how about:

```sh
GOPATH ?= $(shell go env GOPATH)
```

alex-aizman:
Thanks for catching this - pushed a tiny fix with conditional assignment for GOPATH/GOBIN; closing this PR.

PR #307, opened 2026-06-09
authn: add Redis KV backend for multi-replica AuthN

### LINKED ISSUE #273: Support for multi-replica AuthN deployments

Rohithmatham12:
## Summary
- Add generic `auth.db.service` config plus `AIS_AUTHN_KV_*` environment overrides for external AuthN KV services.
- Add a Redis-backed AuthN KV driver selected with `auth.db.type = "Redis"`; local BuntDB remains the default backend.
- Share the same collection/key layout across BuntDB and Redis via `kvdb.MakePath`.
- Add Redis driver coverage with `miniredis` for set/get/delete/list/get-all/delete-collection behavior.
- Document Redis-based shared AuthN storage for multi-replica deployments and add third-party notices for the new Redis-related dependencies.

Closes #273.

## Testing
- `gofmt -w api/env/authn.go api/authn/config.go api/authn/config_internal_test.go cmd/authn/config/cmgr.go cmd/authn/config/cmgr_test.go cmd/authn/kvdb/kvdb.go cmd/authn/kvdb/redis.go cmd/authn/kvdb/redis_internal_test.go cmn/kvdb/api.go cmn/kvdb/bunt.go`
- `go test ./api/authn ./cmd/authn/config ./cmd/authn/kvdb ./cmd/authn/signing ./cmn/kvdb`
- `go test ./cmd/authn`
- `git diff --check`

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,0 +1,54 @@
+AIStore includes the following third-party software with their respective licenses:
+
+--------------------------------------------------------------------------------
+github.com/redis/go-redis/v9
+Copyright (c) 2013 The github.com/redis/go-redis Authors.
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are
+met:
+
+   * Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+   * Redistributions in binary form must reproduce the above
+copyright notice, this list of conditions and the following disclaimer
+in the documentation and/or other materials provided with the
+distribution.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+"AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+--------------------------------------------------------------------------------
+github.com/alicebob/miniredis/v2
+The MIT License (MIT)
+
+Copyright (c) 2014 Harmen
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+--------------------------------------------------------------------------------
--- a/api/authn/config.go
+++ b/api/authn/config.go
@@ -37,7 +37,9 @@ const (
 	defaultPort             = 52001
 	defaultTokenExpiration  = cos.Duration(24 * time.Hour)
 	defaultMaxTokenAge      = cos.Duration(90 * 24 * time.Hour)
-	defaultKVServiceAddr    = "localhost:6379"
+	defaultKVServiceHost    = "localhost"
+	defaultKVServicePort    = 6379
+	defaultKVServiceTimeout = cos.Duration(5 * time.Second)
 )
 
 // Signing key management modes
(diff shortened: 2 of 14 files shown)


api/authn/config.go
```diff
@@ -37,6 +37,7 @@ const (
 	defaultPort             = 52001
 	defaultTokenExpiration  = cos.Duration(24 * time.Hour)
 	defaultMaxTokenAge      = cos.Duration(90 * 24 * time.Hour)
+	defaultKVServiceAddr    = "localhost:6379"
```
aaronnw: Let's split port away from address.

api/env/authn.go
```diff
@@ -26,6 +26,10 @@ const (
 	AisAuthPrivateKeyFile = "AIS_AUTHN_PRIVATE_KEY_FILE"
 	AisAuthPrivateKeyPass = "AIS_AUTHN_PRIVATE_KEY_PASS" // optional passphrase for encrypting private key on disk
 	AisAuthPublicKey      = "AIS_AUTHN_PUBLIC_KEY"       // for asymmetric tokens
+	AisAuthKVAddr         = "AIS_AUTHN_KV_ADDR"          // external KV service address for AuthN DB backends
+	AisAuthKVPassword     = "AIS_AUTHN_KV_PASSWORD"      // optional external KV service password
+	AisAuthKVDBIndex      = "AIS_AUTHN_KV_DB_INDEX"      // optional external KV service database/index
+	AisAuthKVTLS          = "AIS_AUTHN_KV_TLS"           // optional TLS toggle for external KV service
```
aaronnw: I don't think we should support env var overrides for any of these except maybe password. We can expect the config contains everything we need to configure DB access.

api/authn/config.go
```diff
@@ -91,8 +92,15 @@ type (
...
 	DatabaseConf struct {
-		DBType   string `json:"type"`
-		Filepath string `json:"filepath"`
+		DBType   string        `json:"type"`
+		Filepath string        `json:"filepath"`
+		Service  KVServiceConf `json:"service"`
+	}
+	KVServiceConf struct {
+		Addr       string `json:"addr,omitempty"`
+		Password   string `json:"password,omitempty"` //nolint:gosec // configuration field, not a hardcoded credential
```
aaronnw: For password, I think we should not need to put it in config. If we do, we need to use `cmn.Censored` and use `json:"-"` similar to ServerConf.psecret`. 

But I would question if there's any value in having it if we do not persist it. I would expect for real deployments we'll use k8s env from secret or vault injection.

api/authn/config.go
```diff
@@ -190,6 +201,16 @@ func (c *SigningKeyConf) validate() error {
 	return nil
 }
 
+func (c *DatabaseConf) validate() error {
+	if c.Service.DBIndex < 0 {
+		return fmt.Errorf("invalid auth.db.service.db_index=%d (must be >= 0)", c.Service.DBIndex)
+	}
+	if c.DBType == "Redis" && c.Service.Addr == "" {
```
aaronnw: Can this be a typed enum like BuntDB in kvdb.go?

cmd/authn/kvdb/redis.go
```diff
@@ -0,0 +1,182 @@
...
+
+func (r *redisDriver) GetAll(collection, pattern string) (map[string]string, int, error) {
+	pattern = normalizePattern(pattern)
+	fullPattern := kvdb.MakePath(collection, pattern)
+	values := make(map[string]string)
+	code, err := r.scan(fullPattern, func(batch []string) error {
+		for _, path := range batch {
+			val, err := r.client.Get(context.Background(), path).Result()
+			if err != nil {
+				if err == redis.Nil {
```
aaronnw: prefer errors.Is()

cmd/authn/kvdb/redis.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+var _ kvdb.Driver = (*redisDriver)(nil)
+
+func newRedisDriver(conf authn.KVServiceConf) (*redisDriver, error) {
+	opts := &redis.Options{
+		Addr:     conf.Addr,
+		Password: conf.Password,
+		DB:       conf.DBIndex,
+	}
```
aaronnw, line 41: I believe for the cancel context created below to work we need `ContextTimeoutEnabled` in the driver options here.

cmd/authn/kvdb/redis.go
```diff
@@ -0,0 +1,182 @@
...
+	if conf.TLSEnabled {
+		opts.TLSConfig = &tls.Config{MinVersion: tls.VersionTLS12}
+	}
+	client := redis.NewClient(opts)
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
aaronnw: I think this timeout would be a valid addition to `KVServiceConf`

cmd/authn/kvdb/redis.go
```diff
@@ -0,0 +1,182 @@
...
+	defer cancel()
```
aaronnw, line 48: I think all calls below that are using context.Background should instead do a similar wrap, e.g.: 

```go
func (r *redisDriver) SetString(collection, key, data string) (int, error) {
	ctx, cancel := r.ctx()
	defer cancel()
	if err := r.client.Set(ctx, kvdb.MakePath(collection, key), data, 0).Err(); err != nil {
		return http.StatusInternalServerError, err
	}
	return http.StatusOK, nil
}
```
and use a common ctx() func to define a `context.WithTimeout(context.Background(), r.timeout)`

For multi-step operations like GetAll, then re-use the same context across all calls.

aaronnw (changes requested):
Thanks for opening! I left some comment on the implementation, will look at the go mod additions and licensing separately.

Rohithmatham12:
Updated the PR to address the Redis KV review feedback:

- split Redis service configuration into host and port
- removed KV env overrides except AIS_AUTHN_KV_PASSWORD
- kept the Redis password out of persisted config and inject it only at runtime
- changed auth.db.type to typed AuthN DB driver values
- added auth.db.service.timeout and wired Redis operations through per-operation timeout contexts
- enabled Redis ContextTimeoutEnabled and switched Redis nil checks to errors.Is
- updated Redis/config tests and AuthN docs for the new config shape

Validation run locally:

- go test ./api/authn ./cmd/authn/config ./cmd/authn/kvdb ./cmd/authn/signing ./cmn/kvdb
- go test ./cmd/authn
- git diff --check

NOTICE
```diff
@@ -0,0 +1,54 @@
...
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
```
alex-aizman, line 53: we cannot place this text in the repo root 

it is a policy decision, and it is out of scope

cmd/authn/kvdb/kvdb.go
alex-aizman, line 89: there's a single shared Redis with no replication

but when two AuthN servers both write - e.g. via the same unmodified `PersistKeyData` - there's a big problem

it is big enough to reject this entire commit - shows total lack of design

alex-aizman:
Closing (comments inline)

PR #310, opened 2026-06-09
object: add limited CSV select CLI prototype

### LINKED ISSUE #305: S3 Select equivalent: server-side SELECT/predicate pushdown (SelectObjectContent) for CSV/JSON/Parquet
### Describe the feature

Add an S3 Select equivalent — a server-side API to run SQL-style projection + filtering over structured objects (CSV, JSON/JSONL, Parquet) on the AIStore targets, returning only the matching rows/columns instead of the full object.

Ideally exposed two ways:
1. S3-compat: implement the SelectObjectContent S3 API in the s3compat layer (so existing S3 Select clients — AWS SDKs, Spark, DuckDB's httpfs, etc. — work unchanged), and
2.  Native: an ais command, e.g. ais object select s3://bck/obj --query "SELECT a,b WHERE c>10" --input parquet --output jsonl.

Execution should be data-local on the target owning the object (same path as ETL), so only the reduced result set crosses the network.

### Use Case

- We store large structured datasets (Parquet/CSV data-lake exports) on an AIStore. Reading whole objects just to filter a few rows/columns wastes bandwidth and client compute.
- AIStore's ETL already does "bring compute to data," but it requires building and deploying a custom transformer container per query — too heavy for ad-hoc / interactive predicate pushdown.
- A standard S3-Select-compatible API would let existing analytics tooling (Spark, DuckDB, Trino, AWS SDK) push predicates to AIStore with no code changes, enabling "query in place" on AI/analytics datasets.
- 

### Proposed Solution

- Implement SelectObjectContent in s3compat: SQL subset, input CSV/JSON/Parquet, output CSV/JSON, streaming/chunked results, mirroring AWS semantics for client compatibility.
- Reuse the existing ETL / data-local execution path, but embed a built-in DuckDB (or Apache Arrow) "select" runtime so users get column/row pushdown without authoring a transformer.
- For Parquet, leverage row-group/column pruning (predicate + projection pushdown) for efficiency.
- native CLI wrapper: ais object select ... and ais ls-style integration.

### Other Information

_No response_

### Acknowledgements

- [ ] I may be able to implement this feature request

v4.7

Ubuntu 26.04

Rohithmatham12:
## Summary
- Add `ais object select` as a limited CLI prototype for #305.
- Support CSV input/output with SQL-like projection and simple predicates, for example `SELECT name,score WHERE score > 10`.
- Stream object contents through the existing object reader and write selected rows to stdout.
- Add unit coverage for projection, `SELECT *`, string/numeric predicates, unknown columns, and optional `FROM` syntax.

## Scope
This is intentionally a small first patch to show `ais object select` in action. It does not attempt S3 Select compatibility, Parquet, DuckDB embedding, or target-side execution yet. The current implementation uses the existing object GET path and keeps the query/evaluation logic isolated so the next stage can replace the executor with an AIS-native target-side implementation.

## Testing
- `go test ./cli`

Closes #305 only if maintainers decide this prototype is the desired first stage; otherwise it can serve as the discussion patch for stage-one requirements.

### DIFF
--- a/cmd/cli/cli/const.go
+++ b/cmd/cli/cli/const.go
@@ -69,6 +69,7 @@ const (
 	commandPut       = "put"
 	commandRemove    = "rm"
 	commandRename    = "mv"
+	commandSelect    = "select"
 	commandSet       = "set"
 
 	// multipart upload commands
@@ -364,7 +365,8 @@ const (
 	mptCompleteArgument = objectArgument + " UPLOAD_ID PART_NUMBERS"
 	mptAbortArgument    = objectArgument + " UPLOAD_ID"
 
-	getObjectArgument = "BUCKET[/OBJECT_NAME] [OUT_FILE|OUT_DIR|-]"
+	getObjectArgument    = "BUCKET[/OBJECT_NAME] [OUT_FILE|OUT_DIR|-]"
+	selectObjectArgument = objectArgument + " --query QUERY"
 
 	optionalPrefixArgument = "BUCKET[/OBJECT_NAME_or_PREFIX]"
 	putObjectArgument      = "[-|FILE|DIRECTORY[/PATTERN]] " + optionalPrefixArgument
--- a/cmd/cli/cli/object_hdlr.go
+++ b/cmd/cli/cli/object_hdlr.go
@@ -204,6 +204,12 @@ var (
 			forceFlag,
 			encodeObjnameFlag,
 		},
+		commandSelect: {
+			objectSelectQueryFlag,
+			objectSelectInputFormatFlag,
+			objectSelectOutputFormatFlag,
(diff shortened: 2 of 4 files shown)


Nahemah1022:
@Rohithmatham12 thanks for the contribution. I’m going to close this PR because it only implements a client-side CSV filter over a normal full-object GET, which moves in the opposite direction from what we need. The filtering, selection, and predicate-pushdown logic needs to be exercised on the server/target side. What we were asking for was a limited but functional end-to-end MVP that shows `ais object select` running through AIS itself exercising the server/target-side path, so we can discuss stage-one requirements from something real. A CLI-only prototype does not help us evaluate that design. If you’d like to continue, please start with a minimal, functional implementation path first.

PR #311, opened 2026-06-09
copy: retry remote cold-read t2t sends

### LINKED ISSUE #309: copy-listrange (cp from remote S3) aborts on large objects (5–50MB) via inter-target objstream "unexpected EOF"

- [x] I have searched the existing issues

### Describe the bug

 Copying a remote S3 (MinIO) bucket into a native `ais://` bucket via
  `ais cp s3://BCK ais://BCK --all` (the `copy-listrange` multi-object xaction)
  **reliably aborts when it reaches large objects (~5–50 MB)**. The failure is in the
  **inter-target object stream** (`objstream` / SBR): the receiving target reports
  `sbr_obj_data_size … unexpected EOF`, the sender `terminated [unexpected EOF … dropping sobj … size=NNNN]`,
  and the entire xaction aborts.

  **Small objects copy fine** — the problem is specific to *large* objects streamed target→target
  during `copy-listrange`. At low concurrency (`--num-workers 1–4`) a small set of the same large
  objects copies OK; at default (auto ≈224) or `--num-workers 16` under sustained load it aborts.

### Expected Behavior

`copy-listrange` should copy large (5–50 MB) objects between targets reliably — as it does for
  small objects, and as AIStore already does for large *internal* objects (e.g., 50–100 MB archive shards).

### Current Behavior

The xaction aborts with inter-target objstream EOF errors. Verbatim:

  ```
  ErrSBR: tco-XTAfibWnBc[UycHhdPX<=QqskCZap] sbr_obj_data_size:[ctx:
    ais://BCK/cerner/appointment/data/00094-…parquet(size=12664418)
    err: unexpected EOF]: PUT /v1/objstream/tco-XTAfibWnBc (t[QqskCZap]: recv.go:121)

  base:172 tco-XTAfibWnBc[QqskCZap=>UycHhdPX] terminated
    [unexpected EOF ctx: dropping sobj-ais://BCK/cerner/condition/data/00139-…parquet(size=50327906)]

  sendobj:337 unexpected EOF

  # job state:
  copy-listrange  Aborted: "tco-XTAfibWnBc[JYqmQqUW=>QqskCZap] terminated [unexpected EOF]"
  ```

### Steps To Reproduce

1. A native `ais://BCK` with a **remote S3 backend** (community MinIO) holding a mix of small
     plus **large (12–50 MB)** objects (in our case 20.9M objects including ~27K parquet of 12–50 MB).
2. Run `ais cp s3://BCK ais://BCK --all` (or the `copy-listrange` API with default / `--num-workers 16`).
3. When the copy reaches the large objects, the xaction aborts with the `objstream`/SBR `unexpected EOF` above.

### Possible Solution

_No response_

### Additional Information/Context

Controls showing it's specific to `copy-listrange`'s inter-target stream — not MinIO, the network, or the data:
  - **rclone** S3 GET (MinIO) → S3 PUT (AIStore native) of the *same* large parquet (incl. the 50 MB one): **succeeds**.
  - `ais get` of 48 MB objects from the same MinIO: **single + 16 concurrent = 0 failures**.
  - AIStore `archive` created/handled **50–100 MB internal shards** without issue.
  - `iperf3` target↔target = **line rate (40 GbE)**.

  Possibly related (separate issue): `ais cp s3://BCK ais://BCK --all` on the 20.9M-object remote bucket
  **silently copied only ~1,100 objects and reported success with no error** (the remainder were never enumerated)
  — a data-loss-risk completeness issue.

  Backend: community MinIO `RELEASE.2025-07-23`, attached per-bucket via `extra.aws.profile` + `extra.aws.endpoint`.

### AIStore build/version

aisnode v4.7 (4.7.16299b6, build 2026-06-03T19:54:25Z, image `aistorage/aisnode:v4.7-oteltracing`); CLI v1.30.16299b6

Kubernetes (kubeadm 1.36.1), AIS Operator v2.19.0, 3 proxies + 3 targets, bare-metal Linux x86_64;   remote backend = MinIO community RELEASE.2025-07-23

Rohithmatham12:
Summary
- Route remote-backend cold-read copies to another target through the existing per-object target PUT path instead of the long-lived data mover objstream.
- Retry remote cold-read transfers by reopening `GetROC` on retriable mid-stream failures such as `unexpected EOF`, short reads, connection resets, broken pipes, and timeouts.
- Keep normal in-cluster copies on the data mover path.
- Preserve copied byte accounting for reader-backed sends by returning the source object size after successful send/PUT.

Why
For copy-listrange from remote S3/MinIO into an AIS bucket, a remote reader can fail or close after the object header has already been sent over objstream. At that point the receiver is still waiting for the advertised object bytes, so the shared stream reports `sbr_obj_data_size` / `unexpected EOF` and the whole xaction can abort.

The fix keeps remote cold reads off the shared objstream and, importantly, retries the per-object transfer with a freshly opened backend reader when the source stream drops mid-object. That handles the backend-reader failure mode without changing the normal in-cluster DM path.

Testing
- `go test ./ais`
- `git diff --check`

Fixes #309

### DIFF
--- a/ais/tgtobj.go
+++ b/ais/tgtobj.go
@@ -1793,6 +1793,9 @@ func (coi *coi) do(t *target, dm *bundle.DM, lom *core.LOM) (res xs.CoiRes) {
 			}
 			coi.OAH = resp.OAH
 			r = resp.R
+			if resp.Remote && r != nil {
+				return coi.sendRemoteColdRead(t, lom, tsi, resp)
+			}
 		}
 		return coi.send(t, dm, lom, r, tsi) // lom is the source of reader if no reader specified
 	}
@@ -2018,6 +2021,42 @@ func (coi *coi) _chunk(t *target, lom, dst *core.LOM, dstChunkSize int64) (res x
 	return res
 }
 
+func (coi *coi) sendRemoteColdRead(t *target, lom *core.LOM, tsi *meta.Snode, resp core.ReadResp) (res xs.CoiRes) {
+	const attempts = 3
+
+	for attempt := 0; attempt < attempts; attempt++ {
+		if attempt > 0 {
+			resp = coi.GetROC(lom, coi.LatestVer, coi.Sync, coi.ETLArgs)
+			if resp.Err != nil {
+				return xs.CoiRes{Err: resp.Err, Ecode: resp.Ecode}
+			}
+			if !resp.Remote || resp.R == nil {
+				coi.OAH = resp.OAH
+				return coi.send(t, nil, lom, resp.R, tsi)
+			}
+		}
+
+		coi.OAH = resp.OAH
+		res = coi.send(t, nil, lom, resp.R, tsi)
+		if res.Err == nil || attempt == attempts-1 || !isRemoteColdCopyRetryable(res.Err) {
+			return res
+		}
+	}
+	return res
+}
+
+func isRemoteColdCopyRetryable(err error) bool {
+	if err == nil {
+		return false
+	}
+	if errors.Is(err, io.ErrUnexpectedEOF) || errors.Is(err, io.EOF) || cos.IsErrRetriableConn(err) {
+		return true
+	}
+	errmsg := err.Error()
+	return strings.Contains(errmsg, "unexpected EOF") ||
+		(strings.Contains(errmsg, "read") && strings.Contains(errmsg, "shorter than size"))
+}
+
 // send object => designated target
 // * source is a LOM or a reader (that may be reading from remote)
 // * one of the two equivalent transmission mechanisms: PUT or transport Send
@@ -2069,6 +2108,9 @@ func (coi *coi) _send(t *target, lom *core.LOM, sargs *sendArgs) (res xs.CoiRes)
 		res.Err = coi._dm(lom /*for attrs*/, sargs)
 	} else {
 		res.Err = coi.put(t, sargs)
+	}
+	if res.Err == nil && sargs.objAttrs != nil {
+		res.Lsize = sargs.objAttrs.Lsize(true)
 	}
 	return res
 }


Nahemah1022:
@Rohithmatham12 Thanks for opening the PR! Can you confirm if this PR actually resolves the problem in your environment? This patch changes the object transport between targets from using the data mover (DM) to a plain HTTP PUT. However, the backend source reader (from MinIO) could still drop mid-stream, which would likely fail in the exact same way. It would be interesting to know if this approach actually resolved the failures on your end.

Rohithmatham12:
Good catch. I amended the PR so it no longer only switches the transport path.

Remote cold-read copies to another target now:

- stay off the long-lived DM objstream, so a short source read cannot poison the shared inter-target stream;
- retry the per-object target PUT by reopening `GetROC` on retriable mid-stream failures (`unexpected EOF`, short read, reset/pipe/timeout), so a dropped MinIO/S3 reader gets a fresh backend GET instead of replaying a consumed reader;
- keep normal in-cluster copies on the DM path.

That addresses the failure mode you called out: if the backend source reader drops mid-stream, the copy now retries with a fresh remote reader before returning an object error. I validated the focused package test locally:

```text
go test ./ais
ok   github.com/NVIDIA/aistore/ais  38.471s
git diff --check
```

The latest PR head is `ca6d28e5683343799a2ef66864235bc562c868f1`. This should be ready for validation against the reported high-concurrency MinIO copy-listrange setup.

Nahemah1022:
As I mentioned in #309, we do not think blindly retrying `unexpected EOF` is the right fix here.

I’ll close this PR for now since this is not the right direction for the fix, and the underlying issue is technically not an AIS bug. That said, we appreciate your time working on it and would encourage reopening or reworking the PR toward clearer backend-origin error reporting.

For example, the error could explicitly show which backend it originated from and why. That would make this failure mode much easier to diagnose without hiding potentially non-retryable backend errors.

Issue #312, opened 2026-06-10
Client ecosystem


kesavkolla:
### Describe the feature

Currently there are only two SDK (golang and Python) available for AIStore.

Would be good to decouple the SDK to it's own repo so that more and more client SDKs can be developed independently.

Rust and Java are other two popular languages in the area of data processing so SDKs for Java and rust would be beneficial. 

Integrations into OpenDAL, fssspec etc... will increase use of adoption.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AIStore build (latest, v3.22, ...)

latest

### Environment details (OS name and version, etc.)

Ubuntu

pradeep-mj:
Hi @kesavkolla, thanks for opening this, and for the other issues and ideas you have filed recently. To prioritize the client-ecosystem work well, it would really help to understand how you are using or planning to use AIStore.

On the idea itself: growing language coverage, and especially OpenDAL / fsspec
integrations, are a high-leverage way to connect AIS to existing data tooling,
so this is worth exploring. To scope it, a few questions:

1. **Priority**:  of the items you listed (Rust SDK, Java SDK, OpenDAL, fsspec),
   which one or two would unblock you the most?
2. **How you are using AIStore**: to make sure we build the right thing, could
   you say a bit about how AIS fits into your work, or how you plan to use it?
   For example: production, proof-of-concept, or evaluation; what kinds of
   workloads and roughly what scale; which languages and engines are on the
   client side; and where a Rust/Java SDK or an OpenDAL/fsspec layer would
   actually slot in.
3. **Integration surface**: for OpenDAL/fsspec specifically, which engines or
   frameworks would you want to point at AIS? That tells us which API surface to
   get right first.
4. **Involvement**: would you be interested in helping shape or build any of
   this: design input, a reference integration, or a prototype? A first user
   driving the requirements makes a big difference.

If you can share even a rough sketch of your environment and what you are trying
to achieve, that would help us place this against the rest of the work. I would
also be glad to meet and learn more about your use case directly; feel free to
reach me at pmadhavarapu@nvidia.com and we can find a time.

Thanks again for raising this.

Rohithmatham12:
Thanks for opening this. Based on the maintainer feedback on the related docs PR, it sounds like the useful next step is not adding a roadmap document to the repo, but narrowing the motivation and deciding which client/integration would be worth maintaining.

A possible way to scope this:

[...]

PR #314, opened 2026-06-11
python: add read-only fsspec adapter

Rohithmatham12:
Summary
- Add an optional aistore[fsspec] extra and fsspec entry point for ais:// paths.
- Implement a read-only AISFileSystem supporting object listing, metadata lookup, binary reads, range reads, and file-like opens through the existing Python SDK.
- Add focused unit coverage for listing, info, range reads, open(), and unsupported write mode.
- Document the minimal usage in the adapter package README.

Why
Issue #312 asks about client ecosystem expansion and integrations such as fsspec/OpenDAL. Instead of adding a broad roadmap document, this PR provides a small, reviewable fsspec integration that can validate whether this direction is useful before larger SDK/repository decisions.

Testing
- /private/tmp/aistore-fsspec-venv/bin/python -m pytest python/tests/unit/fsspec/test_filesystem.py -q
- /private/tmp/aistore-fsspec-venv/bin/python -m py_compile python/aistore/fsspec/filesystem.py python/tests/unit/fsspec/test_filesystem.py
- /private/tmp/aistore-fsspec-venv/bin/python - <<'PY'
import fsspec
from aistore.fsspec import AISFileSystem
fs = fsspec.filesystem('ais', endpoint='http://ais.example')
assert isinstance(fs, AISFileSystem)
print('fsspec entry point ok')
PY
- git diff --check

Addresses #312

### DIFF
--- a/python/aistore/common_requirements
+++ b/python/aistore/common_requirements
@@ -17,6 +17,7 @@ pytest-xdist>=3.6.1
 typing-extensions>=4.12.2
 webdataset>=0.2.100
 tenacity>=9.0.0
+fsspec>=2024.2.0
 fastapi>=0.109.1
 httpx>=0.28.0
 aiofiles>=23.2.1
--- a/python/aistore/fsspec/README.md
+++ b/python/aistore/fsspec/README.md
@@ -1,0 +1,23 @@
+# AIStore fsspec Adapter
+
+The AIStore fsspec adapter provides read-only access to AIS objects through the
+standard fsspec filesystem interface.
+
+Install the optional dependency:
+
+```console
+pip install "aistore[fsspec]"
+```
+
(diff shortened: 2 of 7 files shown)


gaikwadabhishek:
Hey @Rohithmatham12 ,
Thanks for the contribution. Let me evaluate if we want to support fsspec with the team and get back to you

gaikwadabhishek:
Hey Rohit, we dont plan to support it for now

Rohithmatham12:
Thanks for checking with the team. Understood that fsspec is not something AIS plans to support right now.

I’ll leave this closed and won’t push further on the client-ecosystem integration unless maintainers decide there is demand later. If priorities change, this branch can still serve as a small read-only prototype/reference for the expected surface area.

PR #321, opened 2026-06-22
copy: include backend context in remote read errors

Rohithmatham12:
Summary

- Remove the previous `RemoteSource` / `RemoteSourceError` plumbing and keep the patch minimal.
- Wrap `GetObjReader` failures at the remote-backend boundary with backend/provider, remote bucket/object, operation, status code when available, and the original cause via `%w`.
- Return that wrapped error through the existing copy/ETL path without adding copy-result context or changing reader behavior.
- Keep copy transport/retry behavior unchanged; `RGET` remains only for stats.
- Add live-cluster integration coverage for copy-listrange on a remote bucket missing source object, asserting the xaction error includes backend/provider and remote object context.

Why

This reworks #321 based on maintainer feedback. The patch now only adds essential remote-backend context at the actual `GetObjReader` failure point, avoiding new diagnostic structs and avoiding reader wrappers or broader result-path plumbing.

Refs #309.

Testing

- `GOCACHE=/private/tmp/aistore-backend-gocache GOMODCACHE=/private/tmp/aistore-backend-gomodcache go test ./core -run '^$'`
- `GOCACHE=/private/tmp/aistore-backend-gocache GOMODCACHE=/private/tmp/aistore-backend-gomodcache go test ./xact/xs -run '^$'`
- `GOCACHE=/private/tmp/aistore-backend-gocache GOMODCACHE=/private/tmp/aistore-backend-gomodcache go test ./ais -run '^$'`
- `GOCACHE=/private/tmp/aistore-backend-gocache GOMODCACHE=/private/tmp/aistore-backend-gomodcache go test -c ./ais/test -o /private/tmp/aistore-ais-test.test`
- `git diff --check`

Not fully run locally:

- `go test ./ais/test -run TestCopyMultiObjRemoteReadErrorContext` requires a live AIS cluster with a configured remote bucket. In this local environment, the integration package attempts to reach `http://localhost:8080` and fails with connection refused.

### DIFF
--- a/core/ldp.go
+++ b/core/ldp.go
@@ -6,6 +6,7 @@ package core
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"net/http"
 	"time"
@@ -123,10 +124,23 @@ func (lom *LOM) GetROC(latestVer, sync bool) (resp ReadResp) {
 remote:
 	// GetObjReader and return remote (object) reader and oah for object metadata
 	// (compare w/ T.GetCold)
-	res := T.Backend(bck).GetObjReader(context.Background(), lom, 0, 0)
+	backend := T.Backend(bck)
+	bucket := bck.String()
+	if remoteBck := bck.RemoteBck(); remoteBck != nil {
+		bucket = remoteBck.String()
+	}
+	res := backend.GetObjReader(context.Background(), lom, 0, 0)
+	resp.Remote = true
 
 	if res.Err != nil {
-		resp.Err, resp.Ecode = res.Err, res.ErrCode
+		if res.ErrCode != 0 {
+			resp.Err = fmt.Errorf("remote backend %s GetObjReader failed for %s/%s (status=%d): %w",
+				backend.Provider(), bucket, lom.ObjName, res.ErrCode, res.Err)
+		} else {
+			resp.Err = fmt.Errorf("remote backend %s GetObjReader failed for %s/%s: %w",
+				backend.Provider(), bucket, lom.ObjName, res.Err)
+		}
+		resp.Ecode = res.ErrCode
 		return resp
 	}
 
@@ -137,8 +151,6 @@ remote:
 		Size:     res.Size,
 	}
 	resp.OAH = oah
-	resp.Remote = true
-
 	// [NOTE] ref 6079834
 	// non-trivial limitation: this reader cannot be transmitted to
 	// multiple targets (where we actually rely on real re-opening);
--- a/ais/test/cp_multiobj_test.go
+++ b/ais/test/cp_multiobj_test.go
@@ -7,6 +7,7 @@ package integration_test
 import (
 	"fmt"
 	"math/rand/v2"
+	"strings"
 	"sync/atomic"
 	"testing"
 	"time"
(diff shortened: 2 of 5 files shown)


Nahemah1022:
@Rohithmatham12 thanks, this is closer to the area I was asking about, but I don’t think we should merge it in this form. Propagating an `RGET` boolean is too coarse: it tells us a remote read was involved, but not which backend operation failed, which remote bucket/object was involved, or whether the failure actually came from the backend versus the target-to-target stream after the reader was opened.  
The error context should be produced at the actual failure point and preserve structured details: provider/backend, remote bucket/object, operation, status/code when available, and the original cause. Also, the current test only covers the formatting helper; it does not exercise the copy-listrange remote-backend path in a real/integration setup.

Rohithmatham12:
Thanks for the review. I reworked the branch to avoid the coarse `RGET`-boolean wrapping.

What changed in the latest push:
- backend context is now produced in `core.GetROC`, at the actual `GetObjReader` boundary
- `core.RemoteReadError` carries provider, remote bucket/object, operation, status code when available, and the original cause
- the returned remote reader wraps non-graceful stream read errors, so an `unexpected EOF` from the backend reader keeps backend context
- copy/xaction code no longer relabels errors later based on `RGET`; that flag remains only for stats
- added coverage for structured backend open/read errors and an AIS copy-reader regression that preserves a remote read error through the copy path

This should distinguish backend reader failures from target-to-target transport failures after the reader is opened, while keeping the copy transport/retry behavior unchanged.

Nahemah1022:
I don’t think the reader wrapper is the right abstraction here. The goal is to preserve/report remote source context, not to change the reader behavior. Maybe simplify this by carrying explicit remote source context through the existing copy/transport result path and include that context where the error is reported. That should make the patch smaller and easier to reason about, while still satisfying the diagnostic goal.

Rohithmatham12:
Thanks, that makes sense. I reworked the branch again to remove the reader wrapper abstraction.

Latest push changes:
- `core.ReadResp` now carries explicit `RemoteSrc` context for remote backend reads
- `xs.CoiRes` carries that context through the existing copy/transport result path
- `GetObjReader` failures are still wrapped at the actual backend boundary with operation/status/cause
- later copy/transport failures are reported as `copy` failures with remote-source context, rather than changing reader behavior or claiming every later error is a backend read failure
- `RGET` remains only for stats
- tests now cover the structured `RemoteSourceError` formatter and the AIS copy-reader path preserving `RemoteSrc`

This should be closer to the simpler shape you suggested.

Nahemah1022:
Hi, we only expect this PR to be a minimal patch that wraps the original error with essential information about the target backend, and then returns that wrapped error right back up. The current scope of your PR is still broader than what we intended. Specifically:
- We don't want to introduce any new struct types, such as `RemoteSource` or `RemoteSourceError`. A single, wrapped error-reporting chain should be sufficient here.
- There is still an integration test missing (one that runs against a live cluster rather than just local unit tests).

Rohithmatham12:
Thanks, I reworked the PR to match that narrower scope.

Latest push (`b8d0ed6df`) changes:

- removed the custom `RemoteSource` / `RemoteSourceError` types
- removed the copy-result remote-source plumbing and later copy/transport wrapping
- now only wraps the original `GetObjReader` error at the remote-backend boundary with provider/backend, remote bucket/object, operation, status code when present, and `%w` original cause
- returns that wrapped error through the existing path without changing reader behavior or copy retry/transport behavior
- added `TestCopyMultiObjRemoteReadErrorContext`, a live-cluster integration test for copy-listrange against a remote bucket with a missing source object; it checks that the xaction error includes the backend/provider and remote object context

Validation run locally:

- `go test ./core -run '^$'`
- `go test ./xact/xs -run '^$'`
- `go test ./ais -run '^$'`
- `go test -c ./ais/test -o /private/tmp/aistore-ais-test.test`
- `git diff --check`

I could not run the new live integration test end-to-end locally because this environment has no AIS cluster listening on `localhost:8080`, but the integration test package compiles.

alex-aizman:
> I could not run the new live integration test end-to-end locally because this environment has no AIS cluster listening on `localhost:8080`

​Please take a look at docs/getting_started.md and note:

It is practically impossible to make contributions to this project without running the local playground - which, by the way, is extremely easy to set up.

​In any case, we cannot and shouldn't review patches that change integration tests without them being fully verified. More importantly, it is absolutely impossible for us to review patches that change source code on the server side unless they've been successfully tested in a live local environment.

Rohithmatham12:
Updated the branch again to make the integration test self-contained for the standard local playground.\n\nWhat changed in the latest push (086171d86):\n- TestCopyMultiObjRemoteReadErrorContext no longer depends on BUCKET being an external remote/cloud bucket.\n- The test now requires the attached remote AIS tier, creates its own source bucket directly on tools.RemoteCluster.URL, copies a missing object through the main cluster via ais://@<remote-uuid>/<bucket>, and checks the propagated xaction error for remote backend/GetObjReader/bucket/object context.\n- This should run against the normal local playground shape from scripts/clean_deploy.sh --deployment all / remais, without cloud credentials or an externally configured remote bucket.\n\nValidation I could complete locally after the change:\n- go test ./core -run '^$'\n- go test ./xact/xs -run '^$'\n- go test ./ais -run '^$'\n- go test -c ./ais/test -o /private/tmp/aistore-ais-test.test\n- git diff --check\n\nI also attempted to bring up the local playground from docs/getting_started.md. This Mac environment could build the binaries after setting GOPATH, but the cluster did not stay up: aisnode reported extreme memory pressure and target out-of-space on the local mountpath, then both localhost:8080 and localhost:11080 refused connections. So I am not claiming an end-to-end live pass from this machine. The test is now shaped to be runnable in a normal Linux local playground with the remote tier attached, e.g. scripts/clean_deploy.sh --target-cnt 1 --proxy-cnt 1 --mountpath-cnt 1 --deployment all --remote-target-cnt 1 --remote-alias remais followed by go test ./ais/test -run '^TestCopyMultiObjRemoteReadErrorContext$' -count=1.

core/ldp.go
```diff
@@ -121,18 +122,25 @@ func (lom *LOM) GetROC(latestVer, sync bool) (resp ReadResp) {
...
+	// GetObjReader and return remote (object) reader and oah for object metadata
 	// (compare w/ T.GetCold)
-	res := T.Backend(bck).GetObjReader(context.Background(), lom, 0, 0)
+	backend := T.Backend(bck)
+	bucket := bck.String()
+	if remoteBck := bck.RemoteBck(); remoteBck != nil {
+		bucket = remoteBck.String()
+	}
+	res := backend.GetObjReader(context.Background(), lom, 0, 0)
+	resp.Remote = true
```
alex-aizman, line 133: hmm.. why all this code, change of the internal API structure, redundant error formatting..

how about the very basic `bck.IsRemote()`

alex-aizman:
comment inline

Rohithmatham12:
Thanks for the inline feedback. I reduced the branch to the basic shape you suggested.\n\nLatest push (ee62686b4) now has only one file changed:\n- core/ldp.go wraps the original GetObjReader error with bck.Cname(lom.ObjName) when bck.IsRemote(), preserving the original error via %w.\n- No internal API/result-struct changes remain.\n- No copy/xaction plumbing changes remain.\n- No integration-test changes remain on this branch.\n\nValidation on the rebased minimal branch:\n- go test ./core -run '^$'\n- git diff --check\n\nSince this PR is already closed, I mainly wanted to note that the branch has been reduced to the one-line remote-bucket context approach from your comment.

PR #330, opened 2026-07-06
etl/webserver: fix direct-put response forwarding

chanu1406:
This PR fixes direct-put response forwarding in the Go ETL webserver.

Before this change, `directPut` treated any 200 response without a positive
`Content-Length` as "delivered to target, no content". Chunked/unknown-length
200 responses from the next pipeline stage (e.g. a streaming ETL server) were
silently dropped while still reporting success. In addition, `handleDirectPut`
set response headers after `WriteHeader` (so they were never sent), never
forwarded `Direct-Put-Length`, and leaked response bodies on several paths.

Changes:
- treat only `Content-Length: 0` as the no-content case; forward
  chunked/unknown-length 200 bodies back as content
- set response headers before `WriteHeader`; forward `Direct-Put-Length` on 204
- close direct-put response bodies on all paths (HTTP and websocket)
- align 200-empty-body handling with the Python servers'
  `handle_direct_put_response` for cross-language parity

### DIFF
--- a/ext/etl/webserver/webserver.go
+++ b/ext/etl/webserver/webserver.go
@@ -269,32 +269,35 @@ func (base *etlServerBase) handleWebsocketPipeline(conn *websocket.Conn, transfo
 	firstURL, remainingPipeline := parsePipelineURL(pipeline)
 	dresp, err := base.directPut(firstURL, transformed, size, objPath, remainingPipeline, etlArgs)
 	if err != nil {
-		writer, _ := conn.NextWriter(websocket.TextMessage)
-		writer.Write([]byte(err.Error()))
-		writer.Close()
+		if writer, werr := conn.NextWriter(websocket.TextMessage); werr == nil {
+			writer.Write([]byte(err.Error()))
+			writer.Close()
+		}
 		return err
+	}
+	if dresp.Body != nil {
+		defer dresp.Body.Close()
 	}
 
 	if dresp.StatusCode == http.StatusOK {
 		// from other ETL server, forward the content back
-		writer, _ := conn.NextWriter(websocket.BinaryMessage)
-		if dresp.Size > 0 && dresp.Body != nil {
-			if _, err := io.Copy(writer, dresp.Body); err != nil {
-				return err
-			}
-			dresp.Body.Close()
-		}
-		writer.Close()
-	} else {
-		// from target, no content
-		writer, _ := conn.NextWriter(websocket.TextMessage)
-		_, err = writer.Write([]byte("direct put success"))
+		writer, err := conn.NextWriter(websocket.BinaryMessage)
 		if err != nil {
 			return err
 		}
-		writer.Close()
+		defer writer.Close()
+		_, err = io.Copy(writer, dresp.Body)
+		return err
 	}
-	return nil
+
+	// from target, no content
+	writer, err := conn.NextWriter(websocket.TextMessage)
+	if err != nil {
+		return err
+	}
+	defer writer.Close()
+	_, err = writer.Write([]byte("direct put success"))
+	return err
 }
 
 func (base *etlServerBase) handleDirectPut(w http.ResponseWriter, transformedReader io.ReadCloser, size int64, objPath, directPutURL, etlArgs string) {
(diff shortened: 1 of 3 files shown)


ext/etl/webserver/webserver.go
```diff
@@ -450,21 +456,22 @@ func (base *etlServerBase) directPut(directPutURL string, r io.ReadCloser, size
 	}
 
 	if resp.StatusCode == http.StatusOK {
-		// from other ETL server, forward the content back
-		if resp.ContentLength > 0 {
+		// from target, no content
+		if resp.ContentLength == 0 {
```
gaikwadabhishek, line 468: this isn't exact parity with Python's `handle_direct_put_response`... Python keys on the read body (`if resp.content:`), so a chunked 200 with an empty body becomes 204/"delivered", while here (ContentLength == -1) it forwards as a 0-byte 200 and the upstream target stores an empty object. I think the Go behavior is actually more correct.. Fix the python side or add a note here saying this

ext/etl/webserver/webserver.go
```diff
@@ -278,11 +278,12 @@ func (base *etlServerBase) handleWebsocketPipeline(conn *websocket.Conn, transfo
...
+		if dresp.Body != nil {
+			_, err := io.Copy(writer, dresp.Body)
+			dresp.Body.Close()
+			if err != nil {
 				return err
 			}
-			dresp.Body.Close()
 		}
 		writer.Close()
 	} else {
```
gaikwadabhishek: this branch doesnt close dresp.Body. please check once

gaikwadabhishek (changes requested):
thanks for contributing.. rebase with root main.. it should fix the CI issues as well

ext/etl/webserver/webserver.go
```diff
@@ -278,11 +278,12 @@ func (base *etlServerBase) handleWebsocketPipeline(conn *websocket.Conn, transfo
...
-			if _, err := io.Copy(writer, dresp.Body); err != nil {
+		if dresp.Body != nil {
+			_, err := io.Copy(writer, dresp.Body)
+			dresp.Body.Close()
+			if err != nil {
 				return err
 			}
-			dresp.Body.Close()
 		}
 		writer.Close()
```
Nahemah1022: Good catch. The previous closing behavior was wrong since we missed the early error returns. I’d prefer this cleaner alternative using standard defers once the resource is successfully initialized:
```
writer, err := conn.NextWriter(websocket.BinaryMessage)
if err != nil {
	dresp.Body.Close()
	return err
}
defer writer.Close()
defer dresp.Body.Close()

_, err = io.Copy(writer, dresp.Body)
return err
```

Nahemah1022:
We recently updated the Python ETL CI job configuration on github, which is likely related to why your python ETL tests on this PR are currently failing. Please rebase your branch onto the latest `main` to pick up [this commit](https://github.com/NVIDIA/aistore/commit/856b6a3ec032ff28b87d2214ca5acc2e5f63e9c2) - the jobs should pass successfully after that.

Nahemah1022:
Note that we currently do not have an integration test that builds `ext/etl/webserver` from the current branch and deploys it as a real ETL job. Adding a testing framework for Go-based webservers that can pick up changes from the current branch and run integration tests against the resulting build would be valuable (similar to [python/tests/integration/sdk/test_etl_ops.py](https://github.com/NVIDIA/aistore/blob/main/python/tests/integration/sdk/test_etl_ops.py) for Python-based webservers), but that is a separate topic..

Because this integration coverage is currently missing, for this PR, please manually test the patched code on a running cluster. The test must specifically exercise a non-empty chunked direct-PUT response - the case this change is intended to fix.

The simplest setup is to:
1. Build an [echo ETL server](https://github.com/NVIDIA/aistore/tree/main/ext/etl/webserver#1-create-a-go-etl-webserver) against this PR.
2. Deploy multiple instances on a [local KinD cluster](https://github.com/NVIDIA/aistore/blob/main/deploy/dev/k8s/README.md) and connect them as an ETL pipeline.
3. Ensure an intermediate direct-PUT response is chunked, and confirm that `directPut` receives `resp.ContentLength == -1`.
4. Verify that the final output is complete and matches the input.

Share the result and how you confirmed that the changed `ContentLength == -1` path was actually exercised.

chanu1406:
Thanks for the review @gaikwadabhishek and @Nahemah1022 

I've rebased so the CI should be fixed now. 

- Websocket path: `dresp.Body` is now closed via `defer` right after `directPut` succeeds, so all branches and early returns are covered. `conn.NextWriter` errors are checked too. Also fixed `wrapHTTPError`, which read error response bodies without closing them.
- Added a note at the `ContentLength == 0` branch explaining the difference from Python. And the Go behavior is more correct like you mentioned, I will open a separate PR to align the Python side.
- Added a unit test for that branch (200 with empty body → 204 + `Direct-Put-Length`).


@Nahemah1022 regarding the verification with the KinD cluster:
- Deployed two echo ETLs built from this branch on a local KinD cluster and ran them as a pipeline. The echo transform returns `size = -1`, so its responses are chunked. 
- Ran both an inline transform and an offline bucket-to-bucket copy with an 8 MiB object; output was byte-identical to the input in both cases.

A temporary log line in `directPut` (removed before pushing) confirmed the fixed path was hit in both runs:

    2026/07/07 22:21:56 directPut: response ContentLength = -1
    2026/07/07 22:22:20 directPut: response ContentLength = -1

Before this fix, that response was treated as "no content" and the body was dropped. The 204 + `Direct-Put-Length` branch isn't reachable on a single-target cluster, so that one is covered by the unit tests.

gaikwadabhishek (approved):

PR #333, opened 2026-07-08
xs/lrit: make 429 retry backoff abort-aware

DivyaNarahari97:
## What
`lsoPage` (added in bb6ac0023) retries remote list-objects on 429 with an
escalating backoff — but the backoff uses a plain `time.Sleep`, which is not
interruptible. Aborting a list/range (multi-object) job mid-backoff blocks
until the sleep finishes: up to ~10s for a single attempt and ~26s cumulative
across all retries.
## Fix
Replace `time.Sleep` with a `select` on the parent xaction's abort channel
(the same idiom already used elsewhere in `lrit.go` and other xactions).
On abort, return the pending too-many-requests error — identical to the
behavior of the existing post-call `IsAborted()` checks, just prompt.

### DIFF
--- a/xact/xs/lrit.go
+++ b/xact/xs/lrit.go
@@ -357,7 +357,13 @@ func (r *lrit) lsoPage(bp core.Backend, lsmsg *apc.LsoMsg, lst *cmn.LsoRes) (eco
 	)
 	for retries = 1; retries < remPageRetriesMax; retries++ {
 		sleep = hk.Jitter(sleep+sleep/2, now+int64(sleep))
-		time.Sleep(sleep)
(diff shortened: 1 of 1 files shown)


alex-aizman:
Hi Divya,

Thank you for taking the time! I'm going to close it, though... There's the usual "tradeoff" type argument: `time.NewTimer` is a heap allocation while xaction abort is a relatively rare event.

Moreover, `time.NewTimer` allocates a new channel (timer.C) - not a lightweight primitive. There's a bunch of related overhead that'd be out of scope to discuss.

Finally, the cost of waiting a few extra seconds for the abort does not seem to outweigh any of the above.

Closing.

PR #326, opened 2026-06-26
object: add target-side CSV select MVP

Rohithmatham12:
Summary
- Add `ais object select` as a limited CSV MVP that routes through AIS instead of filtering locally in the CLI.
- Add a `select-obj` object POST action; the proxy redirects to the owning target, and the target reads the source object via `LOM.GetROC` and streams selected CSV rows.
- Add `api.SelectObjectReader` for callers that want the server-side object-select response stream.
- Keep the first scope narrow: CSV input/output, `SELECT` projection, `SELECT *`, optional `FROM`, and simple string/numeric `WHERE` predicates.
- Add shared selector coverage plus a target-side regression that stores a CSV object and exercises the target selection path.

Why
This replaces the closed client-side prototype in #310 with the server/target-side MVP requested in review. The CLI no longer performs a normal full-object GET and filters locally; it POSTs an object action to AIS, and the target performs the read/filter/stream path.

Scope / non-goals
- No S3 Select compatibility yet.
- No Parquet/JSON/DuckDB integration yet.
- No predicate pushdown into remote backend format readers yet.
- This is a stage-one CSV path intended to make the API, proxy routing, target execution, and streaming response shape reviewable.

Testing
- `GOCACHE=/private/tmp/aistore-select-gocache GOMODCACHE=/private/tmp/aistore-select-gomodcache go test ./cmn`
- `GOCACHE=/private/tmp/aistore-select-gocache GOMODCACHE=/private/tmp/aistore-select-gomodcache go test ./api`
- `GOCACHE=/private/tmp/aistore-select-gocache GOMODCACHE=/private/tmp/aistore-select-gomodcache go test ./ais -run TestObjSelectRunsOnTarget`
- `cd cmd/cli && GOCACHE=/private/tmp/aistore-select-gocache GOMODCACHE=/private/tmp/aistore-select-gomodcache go test ./cli`
- `git diff --check`

Replaces the closed client-side prototype #310.
Refs #305.

PR #335, opened 2026-07-09
object: add target-side CSV select MVP

Rohithmatham12:
Summary

- Add `ais object select` as a limited CSV MVP that routes through AIS instead of filtering locally in the CLI.
- Add a `select-obj` object POST action; the proxy redirects to the owning target, and the target reads the source object via `LOM.GetROC` and streams selected CSV rows.
- Add `api.SelectObjectReader` for callers that want the server-side object-select response stream.
- Keep the first scope narrow: CSV input/output, `SELECT` projection, `SELECT *`, optional `FROM`, and simple string/numeric `WHERE` predicates.
- Compile the selector once per object: projection indexes and predicate column indexes are resolved before the row loop, numeric predicate RHS values are parsed once, `SELECT *` avoids building a projection slice, `csv.Reader.ReuseRecord` is enabled, and projected output reuses a single buffer instead of allocating one per output row.

Why

This replaces the closed client-side prototype #310 and the closed server-side draft #326 with the target-side MVP requested in review, while addressing the hot-path feedback from #326.

The CLI no longer performs a normal full-object GET and filters locally; it POSTs an object action to AIS, and the target performs the read/filter/stream path. The per-row path no longer reparses the user-provided numeric predicate value or allocates a fresh projection slice for every matching row.

Scope / non-goals

- No S3 Select compatibility yet.
- No Parquet/JSON/DuckDB integration yet.
- No predicate pushdown into remote backend format readers yet.
- This is a stage-one CSV path intended to make the API, proxy routing, target execution, streaming response shape, and hot-path behavior reviewable.

Testing

- `GOCACHE=/private/tmp/aistore-select-gocache GOMODCACHE=/private/tmp/aistore-select-gomodcache go test ./cmn`
- `GOCACHE=/private/tmp/aistore-select-gocache GOMODCACHE=/private/tmp/aistore-select-gomodcache go test ./api`
- `GOCACHE=/private/tmp/aistore-select-gocache GOMODCACHE=/private/tmp/aistore-select-gomodcache go test ./ais -run TestObjSelectRunsOnTarget`
- `cd cmd/cli && GOCACHE=/private/tmp/aistore-select-gocache GOMODCACHE=/private/tmp/aistore-select-gomodcache go test ./cli`
- `GOCACHE=/private/tmp/aistore-select-gocache GOMODCACHE=/private/tmp/aistore-select-gomodcache go test ./cmn -bench BenchmarkSelectObjectCSVProjectionPredicate -benchmem -run '^$'`
- `git diff --check`

Replaces #326.
Refs #305.

### DIFF
--- a/ais/proxy.go
+++ b/ais/proxy.go
@@ -2047,7 +2047,7 @@ func (p *proxy) httpobjpost(w http.ResponseWriter, r *http.Request, apireq *apiR
 		return
 	}
 	switch msg.Action {
-	case apc.ActRenameObject, apc.ActCheckLock, apc.ActMptUpload, apc.ActMptAbort, apc.ActMptComplete:
+	case apc.ActRenameObject, apc.ActSelectObject, apc.ActCheckLock, apc.ActMptUpload, apc.ActMptAbort, apc.ActMptComplete:
 		apireq.after = 2
 	}
 	if err := p.parseReq(w, r, apireq); err != nil {
@@ -2141,6 +2141,11 @@ func (p *proxy) httpobjpost(w http.ResponseWriter, r *http.Request, apireq *apiR
 		}
 		objName := msg.Name
 		p.redirectAction(w, r, bck, objName, msg)
+	case apc.ActSelectObject:
+		if err := p.checkAccess(w, r, bck, apc.AccessRO); err != nil {
+			return
+		}
+		p.redirectAction(w, r, bck, apireq.items[1], msg)
 	case apc.ActMptUpload, apc.ActMptAbort, apc.ActMptComplete:
 		if err := p.checkAccess(w, r, bck, apc.AccessRW); err != nil {
 			return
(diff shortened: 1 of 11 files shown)


alex-aizman:
Yes, this addresses my two narrow hot-path comments from #326, but the larger issue remains:

* The implementation introduces ad-hoc CSV/SQL parser in `cmn` with only a tiny subset of query behavior.
* The CLI/API integration is missing: the PR adds api.SelectObjectReader, but the CLI does not use it
* Plumbing: a new object action, proxy routing, target-side execution, SDK API, CLI command, parser, and tests as a single “CSV MVP” - all of the above in one fell swoop, one nonrenewable patch.

Last but not the least - there's no attempt to align with S3 SELECT features, let alone any reviewed AIS SELECT design.

This is not reviewable as a drive-by MVP patch.

PR #337, opened 2026-07-14
Allow setting cluster UUID with AIS_CLUSTER_UUID environment variable

eric-higgins-ai:
This adds an `AIS_CLUSTER_UUID` environment variable to allow setting the cluster UUID when bringing up a new cluster. This is necessary because of the case @aaronnw mentioned in https://github.com/NVIDIA/ais-k8s/pull/57 - specifically

> Cluster ID is set on creation and usually lives in that state and is also reflected in the CR. We need to make sure that it updates if recreated, or maybe even persists from the CR.

### DIFF
--- a/ais/earlystart.go
+++ b/ais/earlystart.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"os"
 	"runtime"
 	"time"
 
@@ -413,6 +414,25 @@ func (p *proxy) secondaryStartup(smap *smapX, primaryURLs ...string) error {
 // the cluster is fully started - e.g., a node restarts with a new IP while the
 // cluster is running.
 
+// _resolveCluUUID determines the cluster UUID to use when the primary bootstraps
+// a new or restarted cluster. Precedence:
+//  1. uuid from regpool joins: use the existing cluster's UUID if the node is joining an existing cluster.
+//  2. AIS_CLUSTER_UUID env var: provided by the Kubernetes operator.
+//  3. newClusterUUID: generate a new UUID when one isn't provided.
+func _resolveCluUUID(uuid, created string, p *proxy) (string, string) {
+	if uuid != "" {
+		return uuid, created
+	}
+
+	if envUUID := os.Getenv(env.AisClusterUUID); envUUID != "" && cos.IsValidUUID(envUUID) {
+		nlog.Infof("%s: reusing cluster UUID %q from %s env", p, envUUID, env.AisClusterUUID)
+		return envUUID, _formattedCluCreationTime()
+	} else if envUUID != "" {
+		cos.ExitLogf("invalid UUID in %s environment variable: %s", env.AisClusterUUID, envUUID)
+	}
+	return _newCluUUID()
+}
+
 func (p *proxy) primaryStartup(loadedSmap *smapX, config *cmn.Config, ntargets int, prim primRes) {
 	var (
 		smap          = newSmap()
@@ -521,11 +541,7 @@ func (p *proxy) primaryStartup(loadedSmap *smapX, config *cmn.Config, ntargets i
 			cos.ExitLog(p.String(), "cannot create cluster with no targets,", smap.StringEx())
 		}
 		clone := smap.clone()
-		if uuid == "" {
-			clone.UUID, clone.CreationTime = _newCluUUID()
-		} else {
-			clone.UUID, clone.CreationTime = uuid, created
-		}
+		clone.UUID, clone.CreationTime = _resolveCluUUID(uuid, created, p)
 		clone.Version++
 		p.owner.smap.put(clone)
 		smap = clone
@@ -606,8 +622,12 @@ func (p *proxy) primaryStartup(loadedSmap *smapX, config *cmn.Config, ntargets i
 
 func _newCluUUID() (uuid, creationTime string) {
 	uuid = cos.GenUUID()
-	creationTime = time.Now().UTC().Format(cos.DateTimeSec)
+	creationTime = _formattedCluCreationTime()
 	return
+}
+
+func _formattedCluCreationTime() string {
+	return time.Now().UTC().Format(cos.DateTimeSec)
 }
 
 func (p *proxy) _cluConfig(smap *smapX) (config *globalConfig, err error) {
@@ -1172,8 +1192,8 @@ ret:
 	p.reg.mpl.RUnlock()
 
 	if after.Smap.version() == 0 || !cos.IsValidUUID(after.Smap.UUID) {
-		after.Smap.UUID, after.Smap.CreationTime = _newCluUUID()
-		nlog.Infoln(p.String(), "new cluster UUID:", after.Smap.UUID)
+		after.Smap.UUID, after.Smap.CreationTime = _resolveCluUUID("", "", p)
+		nlog.Infoln(p.String(), "cluster UUID:", after.Smap.UUID)
 		return after.Smap
 	}
 	if before.Smap == after.Smap {
--- a/api/env/ais.go
+++ b/api/env/ais.go
@@ -57,6 +57,9 @@ const (
 	AisK8sPublicDNSMode        = "AIS_PUBLIC_DNS_MODE"
 	AisK8sEnableExternalAccess = "ENABLE_EXTERNAL_ACCESS"
 
+	// Cluster UUID to use on primary startup.
+	AisClusterUUID = "AIS_CLUSTER_UUID"
+
 	// AisK8sPublicDNSMode values
 	PubNetDNSModeIP   string = "IP"
 	PubNetDNSModeNode string = "Node"


ais/earlystart.go
```diff
@@ -413,6 +414,22 @@ func (p *proxy) secondaryStartup(smap *smapX, primaryURLs ...string) error {
...
+//  1. uuid from regpool joins: use the existing cluster's UUID if the node is joining an existing cluster.
+//  2. AIS_CLUSTER_UUID env var: provided by the Kubernetes operator.
+//  3. newClusterUUID: generate a new UUID when one isn't provided.
+func resolveClusterUUID(uuid, created string, p *proxy) (string, string) {
+	if uuid != "" {
+		return uuid, created
+	}
+	if envUUID := os.Getenv(env.AisClusterUUID); cos.IsValidUUID(envUUID) {
+		nlog.Infof("%s: reusing cluster UUID %q from %s env", p, envUUID, env.AisClusterUUID)
+		return envUUID, time.Now().String()
```
alex-aizman: nit: `time.Now().String()` 

must be extracted from `newClusterUUID()` and must be consistent across codebase. 

But, since I started looking - noticed:

```
$ ais show cluster smap --json | grep creation
    "creation_time": "",
$ 
```

Take a look. Empty creation time is a purely informational knob but not to have it - is still a bug.

Must be a single-line fix somewhere around the same place in earlystart.go where we assign UUID.

Thirdly, time.String() is not the best format:

```
2026-07-14 17:16:40.685002878 -0400 EDT m=+0.000011650
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

I'd say, everything ^^^ marked is a junk.

ais/earlystart.go
```diff
@@ -413,6 +414,22 @@ func (p *proxy) secondaryStartup(smap *smapX, primaryURLs ...string) error {
...
+// resolveClusterUUID determines the cluster UUID to use when the primary bootstraps
+// a new or restarted cluster. Precedence:
+//  1. uuid from regpool joins: use the existing cluster's UUID if the node is joining an existing cluster.
+//  2. AIS_CLUSTER_UUID env var: provided by the Kubernetes operator.
+//  3. newClusterUUID: generate a new UUID when one isn't provided.
+func resolveClusterUUID(uuid, created string, p *proxy) (string, string) {
+	if uuid != "" {
+		return uuid, created
+	}
+	if envUUID := os.Getenv(env.AisClusterUUID); cos.IsValidUUID(envUUID) {
```
alex-aizman: you cannot ignore the failure of `cos.IsValidUUID()` - this is terminal

ais/earlystart.go
```diff
@@ -1165,7 +1178,7 @@ ret:
 	p.reg.mpl.RUnlock()
 
 	if after.Smap.version() == 0 || !cos.IsValidUUID(after.Smap.UUID) {
-		after.Smap.UUID, after.Smap.CreationTime = newClusterUUID()
+		after.Smap.UUID, after.Smap.CreationTime = resolveClusterUUID("", "", p)
 		nlog.Infoln(p.String(), "new cluster UUID:", after.Smap.UUID)
```
alex-aizman: strictly speaking, "new cluster UUID:" may not very new anymore

ais/earlystart.go
alex-aizman, line 451: here's likely the fix:
```
+		smap.CreationTime = loadedSmap.CreationTime
```

ais/earlystart.go
alex-aizman: I'd maybe vote for `time.Now().UTC().Format("2006-01-02 15:04:05")`

eric-higgins-ai: I think I addressed this, but let me know if I misunderstood or missed anything. Now I see
```
# ais show cluster smap --json | grep creation
    "creation_time": "2026-07-14 23:20:19",
```

ais/earlystart.go
eric-higgins-ai: I just changed the log to print `"cluster UUID: <UUID>"` (removing the new). Let me know if you had something else in mind

ais/earlystart.go
```diff
@@ -413,6 +414,24 @@ func (p *proxy) secondaryStartup(smap *smapX, primaryURLs ...string) error {
...
+// a new or restarted cluster. Precedence:
+//  1. uuid from regpool joins: use the existing cluster's UUID if the node is joining an existing cluster.
+//  2. AIS_CLUSTER_UUID env var: provided by the Kubernetes operator.
+//  3. newClusterUUID: generate a new UUID when one isn't provided.
+func resolveClusterUUID(uuid, created string, p *proxy) (string, string) {
+	if uuid != "" {
+		return uuid, created
+	}
+	if envUUID := os.Getenv(env.AisClusterUUID); cos.IsValidUUID(envUUID) {
+		nlog.Infof("%s: reusing cluster UUID %q from %s env", p, envUUID, env.AisClusterUUID)
```
alex-aizman, line 428: and what if your new env var is not defined?

we will never start

ais/earlystart.go
```diff
@@ -413,6 +414,24 @@ func (p *proxy) secondaryStartup(smap *smapX, primaryURLs ...string) error {
...
+		return envUUID, formattedClusterCreationTime()
+	} else {
+		cos.ExitLogf("%s: invalid UUID in %s environment variable: %s", env.AisClusterUUID, envUUID)
+	}
```
alex-aizman, line 432: check ExitLogf

ais/earlystart.go
```diff
@@ -413,6 +414,24 @@ func (p *proxy) secondaryStartup(smap *smapX, primaryURLs ...string) error {
...
+//  1. uuid from regpool joins: use the existing cluster's UUID if the node is joining an existing cluster.
+//  2. AIS_CLUSTER_UUID env var: provided by the Kubernetes operator.
+//  3. newClusterUUID: generate a new UUID when one isn't provided.
+func resolveClusterUUID(uuid, created string, p *proxy) (string, string) {
+	if uuid != "" {
+		return uuid, created
+	}
+	if envUUID := os.Getenv(env.AisClusterUUID); cos.IsValidUUID(envUUID) {
+		nlog.Infof("%s: reusing cluster UUID %q from %s env", p, envUUID, env.AisClusterUUID)
+		return envUUID, formattedClusterCreationTime()
```
alex-aizman: nit: ais/clustermap returns a similar error  - maybe it's time to unify

alex-aizman:
generally, to make it more efficient and less back-and-forth:

* always run `make lint`
* ideally, another 5m to run local playground and a smoke test - as per docs/getting_started.md

will catch many little things

ais/earlystart.go
eric-higgins-ai: :facepalm: yes good point

alex-aizman:
as far as the design question of whether we need this env var - I doubt. There's the clear and present danger of breaking all sorts of referential integrity-type relationships. There's the maintenance burden of tracking multiplying env vars that tend to get scaterred all over the place - very much unlike a single piece of cluster configuration.

Speaking of which, take a look at

https://github.com/NVIDIA/aistore/tree/main/cmd/xmeta

alex-aizman:
FYI, since I noticed an unrelated Smap creation-time bug I just went ahead and fixed it separately:

* a0205d8134f768f7dbde4da75ce3aa802b01f52a

ais/earlystart.go
eric-higgins-ai: at a glance, I feel like it doesn't make sense (assuming you're referring to [this](https://github.com/NVIDIA/aistore/blob/a0205d8134f768f7dbde4da75ce3aa802b01f52a/ais/clustermap.go#L274) since we're exiting the process on an invalid UUID here but ais/clustermap returns an error. Let me know if I'm misunderstanding something though

eric-higgins-ai:
@alex-aizman regarding

> as far as the design question of whether we need this env var - I doubt.

are you suggesting that we should pass the cluster UUID through some other method than an env var? Or that having the kubernetes operator generate the cluster UUID and pass it into AIStore is the wrong way to address the problem I mentioned in the PR description?

alex-aizman:
here's my question: what is it that we ultimately are trying to resolve? The `emptyDir` case when:

* the entire state that was previously stored under `configDir` - is gone
* but targets still have (numerous versioned) copies of BMD stored in their respective mountpaths

And then, to deal with this situation we introduce another entity - from the outside - to provide cluster UUID via environment, so that targets agree to join.

Correct me if I'm wrong.

But if that's the case - AIS_CLUSTER_UUID does not match it. It is by definition an externally supplied cluster ID. It does not entail in its own definition everything that I just said above.

Secondly, see my previous comments in re: env vars. 

There must be a solution that is natural and native to AIStore. And it exists. Maybe give it a bit more thought..

eric-higgins-ai:
Yeah, the case you mentioned is correct.

re: @aaronnw 's comment [here](https://github.com/NVIDIA/ais-k8s/pull/57#discussion_r3591435025) - dealing with this case is likely over-engineering so I'll close this PR

alex-aizman:
ex post facto:

> There must be a solution that is natural and native to AIStore.

There's no reason in the world for targets _not_ to store their copy of Smap on a mountpath alongside BMD...

PR #339, opened 2026-07-17
oci: fix 429 wrapped with nil error (nil-deref panic in ErrTooManyRequests.Error)

### LINKED ISSUE #338: OCI backend: 429 wrapped as NewErrTooManyRequests(nil, ...) → nil-deref panic (SIGSEGV) on the prefetch error path; fleet-wide crash loop

- [x] I have searched the existing issues. No match for an `ErrTooManyRequests` panic / OCI backend nil error. Related but distinct: #315 (closed; a different crash on the same prefetch path — `SetCustomKey` map race) and #333 (merged PR; makes the `xs/lrit` list-objects 429 retry backoff abort-aware — that retry loop covers only list-objects pages, not the per-object `prefetch.do` path, and in any case the bug here is the error object itself, not retry policy).

### Describe the bug

When OCI Object Storage returns HTTP 429 (throttling), the OCI backend maps it with a **nil inner error**:

```go
// ais/backend/oci.go:395 (same line at a0205d813, unchanged through current main d30235c21; identical at our build 2d7ef239c)
case http.StatusTooManyRequests:
    errOut = cmn.NewErrTooManyRequests(nil, statusCode)
```

It is the only call site in the tree that passes nil — the sibling backends (`aws.go:878`, `azure.go:167`, `gcp.go:645`) and all other callers (`ais/tgtmpt.go`, `xact/xs/blob_download.go`, `xact/xs/moss.go`) pass a non-nil error (compact/sanitized forms of the SDK error).

The constructor's guard cannot catch this, in debug builds or otherwise: `cos.IsTypedNil` (`cmn/cos/unsafe.go:74-79`) returns `false` for an *untyped* nil interface, so `debug.Assert(!cos.IsTypedNil(nil))` passes — the guard only catches typed-nil, leaving the non-nil invariant unenforced.

The first time anything formats the error — e.g. a prefetch worker recording it via `xact.(*Base).AddErr` — `(*ErrTooManyRequests).Error()` dereferences the nil:

```go
// cmn/err.go:1039-1041 (deref at 1040 on main; 1041 at build 2d7ef239c)
func (e *ErrTooManyRequests) Error() string {
    return e.err.Error()   // e.err == nil → SIGSEGV
}
```

`aisnode` exits (status 2); systemd restarts it; the prefetch client resubmits; OCI is still throttling → the target panics again within seconds. In our deployment this became a **fleet-wide crash loop** (~13–30 s per cycle, systemd restart counters in the 1700s), recurring immediately after every coordinated cluster restart until the prefetch workload was paused.

### Expected Behavior

An OCI 429 should surface as a normal retryable `ErrTooManyRequests` carrying a non-nil inner error — as every other backend behaves. Backend throttling must never crash the target.

### Current Behavior

Target panics. Representative stack trace at build `2d7ef239c`; on `a0205d813` (unchanged through current main `d30235c21`) the equivalent frames are `cmn/err.go:1040`, `cmn/cos/err.go:181`, `xact/base.go:251`, `xact/xs/prefetch.go:205`, `xact/xs/lrit.go:427` / created-by `:199`:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x1827835]

goroutine 4655874 [running]:
github.com/NVIDIA/aistore/cmn.(*ErrTooManyRequests).Error(...)
        cmn/err.go:1041
github.com/NVIDIA/aistore/cmn/cos.(*Errs).Add(...)
        cmn/cos/err.go:172
github.com/NVIDIA/aistore/xact.(*Base).AddErr(...)
        xact/base.go:246
github.com/NVIDIA/aistore/xact/xs.(*prefetch).do(...)
        xact/xs/prefetch.go:202
github.com/NVIDIA/aistore/xact/xs.(*lrworker).run(...)
        xact/xs/lrit.go:378
created by github.com/NVIDIA/aistore/xact/xs.(*lrit).run in goroutine 4629503
        xact/xs/lrit.go:195
```

### Steps to reproduce

1. Cluster with the native OCI backend (`oc://` provider).
2. Drive `prefetch-listrange` (multiple workers, multiple targets) hard enough that OCI Object Storage returns 429 on cold HEAD/GET.
3. The 429 propagates from `ociErrorToAISError` to `prefetch.do`'s error handling → `r.AddErr(err, ...)` → panic.

The prefetch path is just where we hit it: any consumer that stringifies the error crashes the same way — e.g. a 429 on plain `ListObjects` reaches `lsoPage`'s give-up logging (`lrit.go:374` on main), and the API error-response path (`WriteErr` → `herr.init`) formats it as well.

### Environment

- aisnode `4.5.2d7ef239c` (build 2026-05-18), Linux bare-metal, systemd-managed
- 34 proxies / 34 targets, one proxy+target per host, native `oc://` backend to OCI Object Storage

### Root cause / suggested fix

`errIn` — the original OCI SDK error — is in scope and non-nil whenever `statusCode` is 429: the status is derived exclusively from `respOrStatus` (either a literal `int` — only used for a hard-coded 416 — or an `ocios.*Response`'s `RawResponse.StatusCode`), and every response-passing caller sits in an `err != nil` branch. One-line fix, mirroring the adjacent `StatusRequestedRangeNotSatisfiable` case:

```diff
 	case http.StatusTooManyRequests:
-		errOut = cmn.NewErrTooManyRequests(nil, statusCode)
+		errOut = cmn.NewErrTooManyRequests(errIn, statusCode)
```

No matching/retry semantics change: `ErrTooManyRequests` has no `Unwrap()`, and `IsErrTooManyRequests` / status classification match the outer type only. The only behavioral change is `Error()` text: panic → SDK message. (If the raw OCI SDK verbosity is unwanted — it embeds a per-request `opc-request-id`, which also defeats `Errs.Add` dedup — a compacted form along the lines of `_awsErr` would be even better; the minimal fix above is at least consistent with the 416 case in the same switch.)

**Accompanying guard fix** (what would have made this catchable in CI): the constructor's assert currently passes for untyped nil —

```diff
 func NewErrTooManyRequests(err error, status int) *ErrTooManyRequests {
-	debug.Assert(!cos.IsTypedNil(err))
+	debug.Assert(err != nil && !cos.IsTypedNil(err))
 	return &ErrTooManyRequests{err, status}
 }
```

**Separate finding while tracing callers** (will file separately; noting for cross-reference): `ocimpt.go:77/113/158/186` pass `*http.Response` as `respOrStatus`, but `ociErrorToAISError`'s type switch handles only `ocios.*Response` values and `int` — so those four multipart error paths fall through to `statusCode = 0` and silently lose their HTTP status (a 429 there can't even be classified).

brentchang:
**Problem** — When OCI Object Storage returns HTTP 429, `ociErrorToAISError` wraps it as `cmn.NewErrTooManyRequests(nil, statusCode)` — the only nil-passing call site of that constructor (aws/azure/gcp all pass a non-nil error). The first `Error()` call — e.g. `xact.(*Base).AddErr` on the prefetch path, `lsoPage` give-up logging, or `WriteErr` — dereferences the nil and crashes the target with SIGSEGV. Under sustained throttling this becomes a cluster-wide crash loop (we hit this in production: 34 targets, ~15s cycle, restart counters >1700).

Fixes #338.

**Fix** — pass `errIn`, which is non-nil whenever `statusCode` derives from a response (every `ocios.*Response`-passing caller sits in an `err != nil` branch); mirrors the adjacent 416 case. Also tightens the constructor's guard — `cos.IsTypedNil(nil)` returns `false` for untyped nil, so the existing assert couldn't catch this even in debug builds.

**Test** — `TestOCIErrorToAISError429` feeds a 429 `RawResponse` + SDK error through `ociErrorToAISError`; on the unfixed code it reproduces the exact production panic (nil-deref, `addr=0x18`). It is `oci`-tagged like the rest of `oci_internal_test.go` and runs via `go test -tags oci ./ais/backend/` (CI lint compiles it under the `oci` tag; CI test jobs use `-tags debug`).

### DIFF
--- a/ais/backend/oci_internal_test.go
+++ b/ais/backend/oci_internal_test.go
@@ -9,6 +9,8 @@ package backend
 import (
 	"crypto/rand"
 	"crypto/rsa"
+	"errors"
+	"net/http"
 	"strings"
 	"sync"
 	"testing"
@@ -197,5 +199,31 @@ func TestOCIClientCreatesRegionClientOnce(t *testing.T) {
 		if first != client {
 			t.Fatal("expected all goroutines to receive the same cached client instance")
 		}
+	}
+}
+
+// Regression test: a 429 from OCI must yield an ErrTooManyRequests that carries
+// the SDK error. Previously the OCI backend passed nil to NewErrTooManyRequests,
+// so the first Error() call (e.g. xaction AddErr on the prefetch path) panicked
+// with a nil-pointer dereference and crashed the target.
+func TestOCIErrorToAISError429(t *testing.T) {
+	errSDK := errors.New("Error returned by ObjectStorage Service. Http Status Code: 429. Error Code: TooManyRequests.")
+	resp := ocios.GetObjectResponse{
+		RawResponse: &http.Response{StatusCode: http.StatusTooManyRequests},
+	}
+
+	ecode, err := ociErrorToAISError("GetObj", "test-bucket", "test/object", "", errSDK, resp)
+	if ecode != http.StatusTooManyRequests {
+		t.Fatalf("expected status %d, got %d", http.StatusTooManyRequests, ecode)
+	}
+	if err == nil {
+		t.Fatal("expected a non-nil error")
+	}
+	if !cmn.IsErrTooManyRequests(err) {
+		t.Fatalf("expected ErrTooManyRequests, got %T (%v)", err, err)
+	}
+	// must not panic, and must carry the SDK error text
+	if msg := err.Error(); !strings.Contains(msg, "429") {
+		t.Fatalf("error text should carry the SDK message, got %q", msg)
 	}
 }
--- a/ais/backend/oci.go
+++ b/ais/backend/oci.go
@@ -392,7 +392,7 @@ func ociErrorToAISError(op, bucketName, objectPath, byteRange string, errIn erro
 	case http.StatusRequestedRangeNotSatisfiable:
 		errOut = cos.NewErrRangeNotSatisfiable(errIn, []string{byteRange}, 0)
 	case http.StatusTooManyRequests:
-		errOut = cmn.NewErrTooManyRequests(nil, statusCode)
+		errOut = cmn.NewErrTooManyRequests(errIn, statusCode)
 	default:
 		if bucketName == "" {
 			path = objectPath // possibly itself also ""
--- a/cmn/err.go
+++ b/cmn/err.go
@@ -1032,7 +1032,7 @@ func IsErrXactNonIC(err error) bool {
 // ErrTooManyRequests (429, 503)
 
 func NewErrTooManyRequests(err error, status int) *ErrTooManyRequests {
-	debug.Assert(!cos.IsTypedNil(err))
+	debug.Assert(err != nil && !cos.IsTypedNil(err)) // Error() derefs it; IsTypedNil alone lets untyped nil through
 	return &ErrTooManyRequests{err, status}
 }
 


ais/backend/oci_internal_test.go
```diff
@@ -199,3 +201,29 @@ func TestOCIClientCreatesRegionClientOnce(t *testing.T) {
 		}
 	}
 }
+
+// Regression test: a 429 from OCI must yield an ErrTooManyRequests that carries
+// the SDK error. Previously the OCI backend passed nil to NewErrTooManyRequests,
+// so the first Error() call (e.g. xaction AddErr on the prefetch path) panicked
+// with a nil-pointer dereference and crashed the target.
+func TestOCIErrorToAISError429(t *testing.T) {
+	errSDK := errors.New("Error returned by ObjectStorage Service. Http Status Code: 429. Error Code: TooManyRequests.")
```
Nahemah1022, line 210: nit: error strings should not be capitalized or end with punctuation.

Make sure to run `make lint` before pushing your commits. it will catch most of these minor formatting issues automatically

Nahemah1022 (changes requested):

Nahemah1022 (approved):

Issue #340, opened 2026-07-18
Use multipart download when cold getting from S3 bucket


eric-higgins-ai:
### Describe the feature

We are caching some fairly large files in AIStore (~5 GB) stored in an OCI bucket. We're using the "AWS" backend with their S3 compatibility API, since it's more mature than the OCI backend. The issue we have is that, when cold getting these files, fetching the file into AIStore is very slow. It would be nice if AIStore would use multipart downloading for better performance.

I'm aware of the [blob downloader](https://docs.nvidia.com/aistore/blob_downloader), which does download large files in multiple parts. It doesn't quite work for our use case though, as it pushes the decision between single-threaded and multi-threaded downloading onto users. As the cluster administrator, I would like a way to configure a multipart download threshold and part size and have it automatically apply to any applicable cold GET requests.

### Use Case

When cold getting files from an upstream S3 bucket, we can get improved performance using multipart downloading. I want to configure this behavior cluster-wide, so users of AIStore don't have to think about whether to use single- or multi-threaded downloads.

### Proposed Solution

There are 2 possible solutions I have in mind:
* Add cluster-wide configuration to allow switching to the blob downloader for files over a certain size. I think this would be similar to the existing `Ais-Blob-Download` header, just on a cluster level instead of a request level
  * Note that this would require a HEAD request before starting the cold GET to determine the implementation to use
  * I believe we would also need to make an update to the behavior when multiple cold GETs are running in parallel for the same object. Right now, all but the first request will error. This contrasts to the normal cold GET implementation, where parallel requests will wait for the first cold GET to complete.
  * I'm also not sure how scalable this is to fetching many files in parallel. I know that the blob downloader is implemented as an xaction, and creating large numbers of xactions isn't always well-supported ([example](https://github.com/NVIDIA/aistore/issues/336)).
* Update the [AWS backend](https://github.com/NVIDIA/aistore/blob/main/ais/backend/aws.go) implementation to use multipart downloading, similar to the way it's implemented in the [OCI backend](https://github.com/NVIDIA/aistore/blob/main/ais/backend/oci.go)

I'm not sure which is preferred - my main reason for opening this issue is to get opinions on that. I prefer the second one, since it's a smaller and simpler change to make, but the first would avoid duplicating the same multipart download behavior across all backend types.

### Other Information

[...]

PR #346, opened 2026-07-24
Allow multipart downloading when pulling from remote

eric-higgins-ai:
This is just an illustration of my idea for addressing https://github.com/NVIDIA/aistore/issues/340

### DIFF
--- a/ais/target.go
+++ b/ais/target.go
@@ -991,7 +991,38 @@ func (t *target) getObject(w http.ResponseWriter, r *http.Request, dpq *dpq, bck
 			nlog.Warningln("GET", lom.Cname(), "via blob-download["+xid+"]:", err)
 			err = nil
 		}
-		return lom, err
+		if xid != "" || err != nil {
+			return lom, err
+		}
+		// xid == "" && err == nil: fall through to regular GET (object already cached
+		// or a concurrent blob download just completed)
+	// server-side multipart (blob) downloading: when enabled, route cold GETs from
+	// remote backends through the built-in blob-downloader (see docs/blob_downloader.md).
+	// Warm GETs (object already cached) fall through to the regular GET path below.
+	// Range reads and archive requests are excluded - the blob-downloader always
+	// fetches the full object.
+	case bck.IsRemote() && r.Header.Get(cos.HdrRange) == "" && !dpq.isArch() && cmn.GCO.Get().Multipart.Enabled:
+		config := cmn.GCO.Get()
+		msg := apc.BlobMsg{
+			ChunkSize:  int64(config.Multipart.PartMaxSize),
+			NumWorkers: config.Multipart.MaxThreads,
+			LatestVer:  _validateWarmGet(lom, dpq.latestVer),
+		}
+		args := &core.BlobParams{
+			RespWriter: w, // NOTE: make a blocking call
+			Lom:        lom,
+			Msg:        &msg,
+			Parent:     "GET",
+		}
+		xid, _, err := t.blobdl(args, nil /*oa*/, w.Header())
+		if err != nil && xid != "" {
+			nlog.Warningln("GET", lom.Cname(), "via blob-download["+xid+"]:", err)
+			err = nil
+		}
+		// xid == "" && err == nil: object is already cached - fall through to regular GET
+		if xid != "" || err != nil {
+			return lom, err
+		}
 	}
 
 	// GET: regular | archive | range
@@ -1982,6 +2013,12 @@ func (t *target) _blobdl(params *core.BlobParams, oa *cmn.ObjAttrs) (string, *xs
 	xid := cos.GenUUID()
 	rns := xs.RenewBlobDl(xid, params, oa)
 	if rns.Err != nil || rns.IsRunning() { // cmn.IsErrXactUsePrev(rns.Err): single blob-downloader per blob
+		// a blob download for this object is already running — wait for it to
+		// complete, then fall through to the regular GET path (object now cached)
+		if cmn.IsErrXactUsePrev(rns.Err) {
+			<-rns.Entry.Get().ChanAbort()
+			return "", nil, nil
+		}
 		return "", nil, rns.Err
 	}
 
--- a/cmd/aisinit/config.go
+++ b/cmd/aisinit/config.go
@@ -79,6 +79,13 @@ var (
 		Timeout: cos.Duration(time.Hour),
 	}
 
+	defaultMultipart = aiscmn.MultipartConf{
+		PartMaxSize: cos.SizeIEC(256 * cos.MiB),
+		Threshold:   cos.SizeIEC(512 * cos.MiB),
+		MaxThreads:  16,
+		Enabled:     false,
+	}
+
 	defaultEC = aiscmn.ECConf{
 		XactConf:     defaultXconf,
 		Enabled:      false,
@@ -230,6 +237,7 @@ func newDefaultConfig() *aiscmn.ClusterConfig {
 		Net:         defaultNet,
 		FSHC:        defaultFSHC,
 		Downloader:  defaultDownloader,
+		Multipart:   defaultMultipart,
 		EC:          defaultEC,
 		Chunks:      defaultChunks,
 		Keepalive:   defaultKeepalive,
--- a/cmn/config.go
+++ b/cmn/config.go
@@ -132,6 +132,7 @@ type (
 		Periodic    PeriodConf      `json:"periodic" allow:"cluster"`
 		Client      ClientConf      `json:"client"`
 		Downloader  DownloaderConf  `json:"downloader"`
+		Multipart   MultipartConf   `json:"multipart"`
 		Features    feat.Flags      `json:"features,string" allow:"cluster"` // to flip assorted global defaults (see cmn/feat/feat and docs/feat*)
 		Version     int64           `json:"config_version,string"`
 		Versioning  VersionConf     `json:"versioning" allow:"cluster"`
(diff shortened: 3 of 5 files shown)


alex-aizman:
Eric,

Thanks for putting in the effort. As an illustration, this did its job: it helped clarify the direction I’ll take. I don’t yet know the timeline.

The next step will be per-request selection via `api.GetArgs`.

Most of the mechanism already exists: `apc.HdrBlobDownload`, `apc.HdrBlobWorkers`, and `apc.HdrBlobChunk`. To keep the development incremental and consistent with the existing code, I may expose those headers directly through `GetArgs`, or add dedicated typed fields. That remains to be seen.

The advantage is that this can be one self-contained commit, with no additional persistent configuration to validate, propagate, and maintain. The tradeoff is that the caller must know it is about to read a large object.

A subsequent step may add a server-side default: cluster config would provide the default, while bucket properties would inherit and optionally override it.

One additional motivation for that second step is support for third-party clients such as Boto3/Botocore.

The threshold decision belongs at the bucket level. A bucket containing 20GB shards and one containing 100KB JSON objects want opposite defaults, and a single cluster-wide setting cannot express that.

Again, I appreciate the push on this one!

alex-aizman:
PS. Speaking of config, we actually already have it. But let us do part-one first, and then due diligence on the second one.

eric-higgins-ai:
Some questions:

> The next step will be per-request selection via api.GetArgs.

What's still needed here? I think the existing headers already expose the things that we want?

> PS. Speaking of config, we actually already have it

Do you have a pointer to it? I didn't see anything like this when I looked at it

alex-aizman:
> Do you have a pointer to it?

https://github.com/NVIDIA/aistore/blob/main/cmn/api.go#L68

and further:

```
$ git grep "Bprops.*Chunks" ais/t*
ais/tgtobj.go:252:      maxMonoSize := int64(poi.lom.Bprops().Chunks.MaxMonolithicSize)
ais/tgtobj.go:259:              return poi.chunk(int64(poi.lom.Bprops().Chunks.ChunkSize))
ais/tgtobj.go:867:      case goi.lom.Bprops().Chunks.AutoEnabled():
ais/tgtobj.go:1808:     dstMaxMonoSize := dst.Bprops().Chunks.MaxMonolithicSize
ais/tgtobj.go:1831:     case lom.Bprops().Chunks.MaxMonolithicSize != dstMaxMonoSize && lom.Lsize() > int64(dstMaxMonoSize):
ais/tgtobj.go:1833:             res = coi._chunk(t, lom, dst, int64(dst.Bprops().Chunks.ChunkSize))
$ 
```

eric-higgins-ai:
is it really desirable to couple multipart download config to chunking config for uploads? Intuitively I would think people may want different configurations for these.

> The next step will be per-request selection via api.GetArgs.
What's still needed here? I think the existing headers already expose the things that we want?

bump on this question

I'm also happy to implement this if that's helpful - I'd like to get this resolved in a somewhat timely fashion

Nahemah1022:
Hi @eric-higgins-ai, feel free to proceed with part one. Please keep it focused on the Go client API - the server path already exists. Add an optional typed blob-download setting to `api.GetArgs`: non-nil enables blob download.
Translate those values into the existing `HdrBlobDownload`, `HdrBlobChunk`, and `HdrBlobWorkers` headers in the shared `GetArgs` request-building path.
Make sure it handles potential conflicts with values already supplied through `GetArgs.Header`. Add focused tests for defaults, explicit values, and conflicts. Once it’s ready, I’ll review it.

eric-higgins-ai:
@Nahemah1022 it's still not clear to me why we're doing this - why do y'all want `HdrBlobDownload`, `HdrBlobChunk`, and `HdrBlobWorkers` to be specified via query params (?) instead of headers? It doesn't seem really related to my request, which was to configure these headers on the server side

Nahemah1022:
@eric-higgins-ai We’re splitting this into two steps. What I mentioned previously is the first step as a client-controlled `BlobThreshold` in `GetArgs`. I’ve implemented it in commit https://github.com/NVIDIA/aistore/commit/5631038802d4cfb92280c8080593576032158a47

The follow-up is a server-side bucket property that automatically uses blob download for cold GET objects above a configured threshold, without client arguments. That aligns with your PR’s goal, but has additional caveats, so we’ll handle it soon and keep this thread updated.

wendy-ouyang-ai:
Hi @Nahemah1022, any update on the server-side follow-up (bucket property that auto-routes cold GETs above a threshold through blob download)?

Nahemah1022:
@wendy-ouyang-ai Thank you for your patience and sorry for the delayed update. We were wrapped up with high-priority items for the previous 5.0 release. We've resumed work on this topic now. The goal is to add a bucket property that automatically routes coldGet through the blob download path when objects exceed a configured size threshold. This feature is planned for the upcoming 5.1 minor release shortly.

wendy-ouyang-ai:
> @wendy-ouyang-ai Thank you for your patience and sorry for the delayed update. We were wrapped up with high-priority items for the previous 5.0 release. We've resumed work on this topic now. The goal is to add a bucket property that automatically routes coldGet through the blob download path when objects exceed a configured size threshold. This feature is planned for the upcoming 5.1 minor release shortly.

Thanks! Do you have a rough ETA? Once the change lands on main, we can cherry-pick the commit onto our deployment without waiting for the full 5.1 release.

Nahemah1022:
@wendy-ouyang-ai We're aiming to land the initial implementation on `main` within the next few days. Note that this feature is planned in two phases:

- Serial mode: The blob-download job persists the entire object and its chunks in AIS first before responding to the client. The client waits for the full download to finish before receiving the first byte. This phase is straightforward and should be ready in the next couple of days.
- Streaming mode: AIS streams individual chunk data back to the client as soon as they land in the cluster. Because this requires additional memory management overhead, it will take a bit longer to complete.

If serial mode fits your workflow, you should be able to cherry-pick the initial commit soon!

wendy-ouyang-ai:
> @wendy-ouyang-ai We're aiming to land the initial implementation on `main` within the next few days. Note that this feature is planned in two phases:
> 
> * Serial mode: The blob-download job persists the entire object and its chunks in AIS first before responding to the client. The client waits for the full download to finish before receiving the first byte. This phase is straightforward and should be ready in the next couple of days.
> * Streaming mode: AIS streams individual chunk data back to the client as soon as they land in the cluster. Because this requires additional memory management overhead, it will take a bit longer to complete.
> 
> If serial mode fits your workflow, you should be able to cherry-pick the initial commit soon!

Serial mode works for us. We don't care too much about first-byte arrival time. Please let us know once the commit is ready to cherry-pick. Thanks @Nahemah1022 !

Nahemah1022:
@wendy-ouyang-ai Thanks for the patience! We're actively working through the implementation now and expect to have it landed shortly, ideally by the end of this week. I'll drop an update here as soon as the PR is merged.

wendy-ouyang-ai:
Hi @Nahemah1022, just want to follow up on the progress

Nahemah1022:
Hi @wendy-ouyang-ai, we are actively working on this, but while working on this feature, we uncovered a few blockers around object locking and concurrency handling that prevent us from cleanly wiring the blob downloader using bucket-level configs just yet. Resolving this properly requires some additional redesign and consolidation of the underlying code.

In the meantime, as a workaround, you can specify the `Ais-Blob-Download: true` header on individual GET requests to engage the blob downloader. We've made several stabilization improvements to this header-triggered path since commit [da7b564](https://github.com/NVIDIA/aistore/commit/da7b56436bad1d68986d0ffe1b0c74e53cfdc78f?utm_source=gemini).

Adding that header to your client's GET requests should allow you to leverage blob downloading right away. Let us know if this helps unblock your workflow.

Thanks again for your patience! We will update you here as soon as the PR for the bucket-triggered path lands.

wendy-ouyang-ai:
Hi @Nahemah1022, I just want to make sure we're describing our goal clearly.  What we were hoping for in [#340](https://github.com/NVIDIA/aistore/issues/340) is server-side behavior where aistore picks the blob downloader based on object size against a threshold. That way users don't need to know whether a dataset has large files, and they don't need an AIS-specific client.

Also, our workloads read through the S3-compatible endpoint from boto3 and Lance, so the GetArgs and BlobThreshold additions on the Go SDK don't reach our readers, and a per-request header is hard for us to apply broadly for the same reason.

wendy-ouyang-ai:
Also, if I understand it correctly, `Ais-Blob-Download: true` and the planned bucket property both trigger the same blob download path on the target, so the locking and concurrency concerns you mentioned would seem to apply to the header as well. I might be missing something, so could you help me understand what makes the header a safe workaround today while the bucket-level trigger still needs the redesign?

Nahemah1022:
@wendy-ouyang-ai Yes, we understand the goal: automatic, server-side selection based on object size and bucket configuration.
You’re also right that both modes use the same underlying blob downloader. The distinction is in how the surrounding GET handles contention and failures, not that the header inherently makes downloading safer. Let me decouple the two:

- Explicit header: the client requests with blob download headers. If the object is write-locked or hits a system error that aborts or prevents the blob downloader from starting, we return that error directly to the client in most cases.
- Implicit bucket policy: blob download is a transparent cold-GET optimization. It must preserve regular GET behavior and allow fallback for eligible failures. That means a failed blob download must safely hand off its object lock to the fallback path. The current codebase doesn't easily support this lock transfer, which is why working through this path is taking longer than expected.

To answer your question: what makes the header a safe workaround is that it's an user-triggered explicit operation, so we don't have to worry too much about fallback logic or the associated locking implications.

I understand adding headers to your client code might be a constraint. However, if your goal is to ensure every qualifying object always be stored chunked, even when blob download fails, the header-triggered path might actually be a better fit. Unlike the bucket-triggered path, it won't silently fall back to the monolithic cold GET behavior without chunking the object. Those are different guarantees.

Could you clarify if this fits your use case? Note that the header-triggered path will still take precedence over the implicit bucket setting in the future, so there won't be any conflict once the bucket config lands.

PR #349, opened 2026-07-29
object: serve and cache cold range reads from remote backend (Range-Cold-GET)

### LINKED ISSUE #348: Ranged GET of a non-cached object downloads the entire object
### Describe the feature

GET with a `Range` header for an object that is not (yet) in the cluster runs the regular cold GET: the target fetches the whole object from the remote backend (`GetObjReader(ctx, lom, 0, 0)`), finalizes it, and only then serves the requested range from the local copy. `Streaming-Cold-GET` doesn't change that - it explicitly bails out on range requests.

For sparse reads over large remote objects this is expensive both ways: time-to-first-byte scales with the object size instead of the range size, and the cluster spends disk and backend bandwidth on data nobody asked for. Reading a 1 MiB slice of a 10 GiB object pulls the full 10 GiB on first access.

The building blocks already exist: `Backend.GetObjReader(ctx, lom, offset, length)` issues real range requests on every backend (S3 `Range` header, GCS `NewRangeReader`, Azure, OCI), and the blob downloader already uses it that way. It's just not wired into the GET path.

### Use Case

Random access over large objects in cloud buckets (seeking readers, index lookups, partial dataset reads) where whole-object caching or prefetching isn't economical.

### Proposed Solution

A feature flag (cluster- and bucket-scope, alongside `Streaming-Cold-GET`): when set and a ranged GET misses, the target HEADs the remote object (the size is needed to resolve open-ended and suffix ranges and to fill `Content-Range`), reads only the requested bytes via `GetObjReader(offset, length)`, and streams them back with 206. The read stays under rlock, so concurrent range reads of the same missing object don't serialize on the cold-GET write lock. Defaults, full GETs and warm reads stay as-is.

Cache what was fetched, so that repeated reads don't keep going to the backend: map the object onto a fixed grid of chunks, store the chunks covering the requested range in a partial chunk-manifest (`core.Ufest`), serve later reads from whatever has accumulated, and complete the manifest once the grid is dense - at which point the object becomes a regular chunked object. The chunked-object machinery already supports this shape: a partial manifest persists, takes chunks out of order and tolerates gaps, while the density and size invariants are only enforced on completion. A fixed grid keeps chunk number and object offset a bijection so gaps need no extra metadata, and deriving the manifest ID from the remote version/ETag keeps chunks from two different versions from ever being stitched together.

Implemented in #349.

### Other Information

Trade-offs the flag opts into: a small range pulls its whole grid chunk when caching, completing the manifest re-reads the local chunks once to compute the whole-object checksum, and an object that is only ever read in part keeps chunks around until `ais space-cleanup` reclaims them. Caching is skipped, though the range is still served, when the remote object has no version/ETag or when `validate_cold_get` is enabled, since a ranged read carries no whole-object checksum to validate.

### Acknowledgements

latest main (5087411)

n/a - applies to any deployment with remote backends

ValentaTomas:
Fixes #348.

A ranged GET of an object that is not in the cluster currently runs the regular cold GET: the whole object gets downloaded and finalized before the range is served from the local copy, so time-to-first-byte scales with the object size and the cluster pays disk and egress for data nobody asked for.

With the new `Range-Cold-GET` flag (cluster- and bucket-scope) the target HEADs the remote object, reads only the requested bytes via the backend's ranged `GetObjReader`, and streams them back with 206 - staying under rlock, so concurrent range reads of the same missing object don't serialize on the cold-GET write lock. Clients see the usual range semantics either way: the same `Range` request and 206/`Content-Range` response, on both the native and `/s3` routes; the flag only changes what the cluster does behind it.

The bytes are cached on the way out. The object is mapped onto a fixed grid (8MiB, coarser for objects that would otherwise exceed `MaxChunkCount` chunks); the chunks covering the range are stored in a partial chunk-manifest (`core.Ufest`), later reads use whatever has accumulated, and once the grid is dense the manifest is completed and the object becomes a regular chunked object. This needs no changes to the chunked-object machinery: a partial manifest already persists, already takes chunks out of order and already tolerates gaps, and the density and size invariants are only enforced on completion. The manifest ID is derived from the remote version/ETag, so chunks from two versions can never be mixed.

Hence `perf,integrity-`, same as `Streaming-Cold-GET`: a small range pulls its whole grid chunk when caching, and completion re-reads the local chunks once to compute the whole-object checksum. Caching is skipped, though the range is still served, when the remote object has no version/ETag or `validate_cold_get` is enabled.

Tests in `ais/test/range_coldget_test.go` (gated on a remote AIS cluster, which `test-short` deploys) plus unit tests for the grid math and range splitting. Docs updated in `docs/feature_flags.md`.

### DIFF
--- a/ais/backend/ht.go
+++ b/ais/backend/ht.go
@@ -190,7 +190,8 @@ func (htbp *htbp) GetObjReader(ctx context.Context, lom *core.LOM, offset, lengt
 	if res.Err != nil {
 		return res
 	}
-	if resp.StatusCode != http.StatusOK {
+	ok := resp.StatusCode == http.StatusOK || (length > 0 && resp.StatusCode == http.StatusPartialContent)
+	if !ok {
 		res.ErrCode = resp.StatusCode
 		res.Err = fmt.Errorf("error occurred: %v", resp.StatusCode)
 		return res
--- a/ais/test/range_coldget_test.go
+++ b/ais/test/range_coldget_test.go
@@ -1,0 +1,272 @@
+// Package integration_test.
+/*
+ * Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
+ */
+package integration_test
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"net/http"
+	"strconv"
+	"strings"
+	"sync"
+	"testing"
+
+	"github.com/NVIDIA/aistore/api"
+	"github.com/NVIDIA/aistore/api/apc"
+	"github.com/NVIDIA/aistore/cmn"
+	"github.com/NVIDIA/aistore/cmn/cos"
+	"github.com/NVIDIA/aistore/cmn/feat"
+	"github.com/NVIDIA/aistore/tools"
+	"github.com/NVIDIA/aistore/tools/readers"
+	"github.com/NVIDIA/aistore/tools/tassert"
+	"github.com/NVIDIA/aistore/tools/tlog"
+	"github.com/NVIDIA/aistore/tools/trand"
+)
+
+// feat.RangeColdGET (see ais/tgtfrange.go): a ranged GET of a not-cached object is
+// served directly from the remote backend, and the grid-aligned chunks covering the
+// requested range get cached; once the grid is fully covered the object becomes
+// a regular (chunked) in-cluster object.
+//
+// NOTE: using a remote AIS cluster ("remais") and out-of-band PUTs, so that the
+// objects in question are, initially, not present in this cluster.
+
+const (
+	rcGridSize = 8 * cos.MiB            // must be in sync w/ rcChunkSize (ais/tgtfrange.go)
+	rcObjSize  = 2*rcGridSize + cos.MiB // 3 grid chunks: two full and one partial
+)
+
+// accumulate the grid, one range read at a time
+func TestRangeColdGet(t *testing.T) {
+	var (
+		proxyURL   = tools.RandomProxyURL(t)
+		baseParams = tools.BaseAPIParams(proxyURL)
+		bck        = rcBucket(t, proxyURL)
+		objName    = "range-cold-get-" + trand.String(5)
+		src        = rcPutRemote(t, bck, objName)
+		m          = &ioContext{t: t, bck: bck}
+	)
+	initMountpaths(t, proxyURL)
+	rcCheckPresent(t, proxyURL, bck, objName, false)
+
+	tlog.Logln("1. read the first chunk")
+	rcGet(t, baseParams, bck, objName, cmn.MakeRangeHdr(0, cos.KiB), src[:cos.KiB])
+	rcCheckPresent(t, proxyURL, bck, objName, false)
+	m.validateChunksOnDisk(bck, objName, -1) // at least one chunk file
+
+	tlog.Logln("2. read the last (partial) chunk - out of order")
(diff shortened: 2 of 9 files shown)


ValentaTomas:
Drafted the implementation here to illustrate a possible solution.
@alex-aizman do you think this is something (at least conceptually) that would make sense to work on and merge here?

alex-aizman:
Tomas, sorry for the delayed response - I’ve been tied up over the past days. And thanks(!) for putting together the draft implementation. 

**1.** I agree with the motivation: directly serving a cold range from the backend while at the same time storing it in AIS, so that - when all set and done - the corresponding remote object exists in chunks and is not _cold_ anymore.

> Separately, I'm saying "storing" not "caching" - with LRU (by default) disabled, we are storing remote data quite persistently.

**2.** Anyway, part of the motivation is right but I'd argue with the following point you make:

>  the whole object gets downloaded and finalized <snip> and the cluster pays disk and egress for data nobody asked for

Not yet (asked). Depends on the app/workflow/access-pattern but largely, most of the time there'll be another range read, and another one after that. So yes, the first range-reader pays the price - but only once. And the benefit is, there's a certain transactional clarity and operational purity in the way it is currently done.

**3.** Now, as far as your implementation, there's immediately a list of what-ifs that, IMHO, make it look fragile and resource-expensive simultaneously.

In your code backend read on the one hand, and client ResponseWriter on the other - are synchronously coupled. This is dangerous. A client reading one byte per second throttles the entire pipeline while the request holds the LOM r-lock:

```
sw = &subrangeWriter{w: goi.w, skip: s.off - s.coff, left: s.size}
...
written, cksum, err := cos.CopyAndChecksum(cos.NewWriterMulti(sw, fh), io.LimitReader(src.r, size), buf, cksumType)
```
Multiple network calls under the one read lock. The slow reader exclusively holds its slot, while the entire pipeline is its hostage.

The reader that, by the way, can just stop reading at any point without dropping the connection - TCP RST won't rescue us, and so we just hang for a while with multiple per-range connections (client <=> AIS and AIS <=> backend), all the associated buffers, and the total overhead.

> The pathological case is N concurrent readers all pulling from remote, none of them caching anything, because one stalled reader holds the slot.

Not to mention numerous externally-concurrent scenarios: PUT(this same object), GET(this same object), GET-range(same range of the same object), and also bucket-level jobs: prefetch, ETL, etc.

**4**. But no - I'm not saying this cannot be done. This can be done of course. Not sure this needs to be done right now - that's a different question. But technically, this is a new `xaction` with a partial manifest under itself, with a well-defined invalidation, admission control (as in: `load.Advice`), internal and Prometheus metrics, well-defined cancelation, and all the rest usual 9-yards - stress-integration testing, docs, comparison (with  vs without) benches, CLI, etc.

alex-aizman:
**5**. My guess is that you had a certain workload in mind when you were thinking about this PR - so maybe (and absolutely optionally) you can share it?

Typical object and range sizes, backend latency, reader concurrency, and/or range overlap/reuse. In particular, what fraction of an object does a typical reader actually touch? How often do readers stall or disconnect mid-range, and what would be expected from AIS when that happens? And separately, how typical (or popular) this type workload could be - in your estimation?

ValentaTomas:
Hey @alex-aizman, thanks for answering.

Yes, I have a specific work in mind and the range read is essentially the primary way for reading data there (adding a [link](https://github.com/e2b-dev/infra/tree/main/packages/shared/pkg/storage) to at least give you some context if you want to check the related code).
I should be able to share more details in the near future, in the meantime I'll close the PR draft and reopen when I can follow up further.

PR #357, opened 2026-08-11
k8s: resolve node name from MY_NODE, drop the startup Pod query

micah-yong-ai:
On startup, `k8s.Init()` queries the API server for the node's own Pod (`client.Pod(podName)`) and calls `cos.ExitLogf` on any error. The only piece of information consumed from the returned Pod spec is `pod.Spec.NodeName`.

This makes a transient API-server connectivity loss at startup fatal. We have observed this at scale: a roughly one-minute network blip on a node (kube API unreachable, `dial tcp 10.96.0.1:443: connect: network is unreachable`) turned into hundreds of aisnode pod terminations, with targets crash-looping under `CrashLoopBackOff` until the network recovered — extending a brief network hiccup into a considerably longer storage outage.

That query is redundant: the node name is already injected into every AIS pod via the K8s downward API.

- the ais-k8s operator sets `MY_NODE` for both targets and proxies — [`operator/internal/resources/aistore/cmn/env.go#L39`](https://github.com/NVIDIA/ais-k8s/blob/ecf9f9859a8ba1468595e6fbc9d669e41d1f317d/operator/internal/resources/aistore/cmn/env.go#L39)
- `deploy/dev/k8s` does the same for local dev, in both the [proxy](https://github.com/NVIDIA/aistore/blob/3f6cf585352c78b041cc90ceda5e517b754f0d4e/deploy/dev/k8s/base/proxy/statefulset.yaml#L98) and [target](https://github.com/NVIDIA/aistore/blob/3f6cf585352c78b041cc90ceda5e517b754f0d4e/deploy/dev/k8s/base/target/statefulset.yaml#L108) statefulsets

So there is no valid in-cluster deployment of AIS without it.

### Changes

- Resolve the node name from `MY_NODE` and drop the startup Pod query entirely — `aisnode` no longer depends on API-server reachability to start.
- Exit with a clean error when `MY_NODE` is not set.
- Extract `_initNode()` from `Init()` (behavior-preserving) to make node-name resolution unit-testable, and add tests covering the three paths: env-provided node name, pod name taken from `HOSTNAME`, and the non-K8s deployment path.

### Testing

With Go 1.26:

- `go build ./...` and `go vet ./cmn/... ./ais/... ./ext/...` — clean
- `gofmt -l cmn/k8s/` — clean
- `golangci-lint run ./cmn/k8s/...` (v2.12.2) — 0 issues
- `go test ./cmn/k8s/...` — passes; the new tests need no K8s cluster

### DIFF
--- a/cmn/k8s/k8s.go
+++ b/cmn/k8s/k8s.go
@@ -1,6 +1,6 @@
 // Package k8s: initialization, client, and misc. helpers
 /*
- * Copyright (c) 2018-2025, NVIDIA CORPORATION. All rights reserved.
+ * Copyright (c) 2018-2026, NVIDIA CORPORATION. All rights reserved.
  */
 package k8s
 
@@ -14,8 +14,6 @@ import (
 	"github.com/NVIDIA/aistore/cmn/cos"
 	"github.com/NVIDIA/aistore/cmn/debug"
 	"github.com/NVIDIA/aistore/cmn/nlog"
-
-	v1 "k8s.io/api/core/v1"
 )
 
 type PodStatus struct {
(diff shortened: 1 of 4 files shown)


micah-yong-ai:
Hi @alex-aizman, I'm Micah on the ML Platform team from Applied Intuition. We recently hit an issue where a brief node-level network blip caused ais-node to crash-loop at startup until networking recovered (a ~1-minute blip turned into several minutes of downtime per target under CrashLoopBackOff).

However, as I took a look at the ais source code, I found that the only load-bearing variable is `pod.Spec.NodeName` and the `ais-k8s` operator already injects that into every pod as `MY_NODE` (i.e., ais-node did not need to fatal in the first place) 

This PR is a candidate fix to fallback to `MY_NODE` with a warning when the pod query fails. Let me know what you think!

cmn/k8s/k8s.go
```diff
@@ -73,9 +73,15 @@ func Init() {
 	}
 
 	// Check Pod.
-	pod, err = client.Pod(podName)
+	pod, err := client.Pod(podName)
 	if err != nil {
-		cos.ExitLogf("Failed to get Pod %q, err: %v", podName, err)
+		if nodeName := os.Getenv(env.AisK8sNode); nodeName != "" {
```
aaronnw: This is good, but I think we can take it one step further. 

The K8s operator always sets "MY_NODE" for both targets and proxies from the k8s downward API. https://github.com/NVIDIA/ais-k8s/blob/ecf9f9859a8ba1468595e6fbc9d669e41d1f317d/operator/internal/resources/aistore/cmn/env.go#L39 

For local dev k8s, we also set it in the statefulset definition: https://github.com/NVIDIA/aistore/blob/3f6cf585352c78b041cc90ceda5e517b754f0d4e/deploy/dev/k8s/base/proxy/statefulset.yaml#L98

So there's no valid deployment of AIS in K8s that doesn't already include this env var. I think this is worth a clean error and a hard exit with no need to query the api server.

aaronnw (changes requested):

cmn/k8s/k8s.go
micah-yong-ai: Awesome, thanks for the context. Just modified to always use `MY_NODE`, and hard error otherwise

micah-yong-ai:
@aaronnw Thanks again for the initial review. Bumping for review whenever you get the chance!

cmn/k8s/k8s.go
```diff
@@ -47,16 +45,16 @@ var (
...
 	}
+	_initNode()
+}
 
-	var (
-		pod     *v1.Pod
-		podName = os.Getenv(env.AisK8sPod)
-	)
+// Resolve this node's name from the environment.
+func _initNode() {
```
aaronnw, line 62: This function is _initNode but it's also getting pod name and logging details. 

Since we no longer use the pod name for resolving node name, then this should be two tasks (separate functions or part of Init()): 

1. validate `MY_NODE` and set to NodeName
2. validate `MY_POD` or `Hostname` and log pod info

cmn/k8s/k8s.go
```diff
@@ -47,62 +45,40 @@ var (
...
-		NodeName = pod.Spec.NodeName
-	case os.Getenv(env.AisK8sNode) != "":
-		NodeName = os.Getenv(env.AisK8sNode)
-	default:
-		cos.ExitLogf("Failed to get K8s node name. %q is not set", env.AisK8sNode)
-		return
+// Resolve this node's name from the environment.
+// MY_NODE is injected via the downward API by every supported K8s deployment
+// (the ais-k8s operator and deploy/dev/k8s alike), so there is no need to ask
+// the API server for our own Pod.
```
aaronnw: Please remove the last couple lines here. I would prefer we don't enumerate all possible deployment types here as that's prone to change. And we don't want to document removed behavior like the api query.

aaronnw: mentioning the downward API is a good change though 👍

micah-yong-ai: Agreed, sorry I let this one slip my review. Just addressed

aaronnw (approved):

PR #358, opened 2026-08-22
ec: fold data checksum into same reedsolomon encode pass

ferhatelmas:
* remove one full source scan
* keep same padding and parity slice logic
* publish slice checksums only after successful encode
* drop slice math from ctx for local ownership and keep allocation out of replica copy path
* preserve raw disk writers for reliable cleanup
* 512 MiB D4 P2 warm-cache runs roughly 2x faster in cksum + encode

### DIFF
--- a/ec/putjogger_test.go
+++ b/ec/putjogger_test.go
@@ -1,0 +1,153 @@
+// Package ec provides erasure coding (EC) based data protection for AIStore.
+/*
+ * Copyright (c) 2018-2026, NVIDIA CORPORATION. All rights reserved.
+ */
+package ec //nolint:testpackage // Tests unexported EC encoding helpers.
+
+import (
+	"bytes"
+	"io"
+	"testing"
+
+	"github.com/NVIDIA/aistore/cmn/cos"
+	"github.com/NVIDIA/aistore/core"
+)
+
+func TestFinalizeSlicesChecksumsPaddedData(t *testing.T) {
+	for _, cksumType := range []string{cos.ChecksumCesXxh, cos.ChecksumSHA256} {
+		t.Run(cksumType, func(t *testing.T) {
+			ctx := newInitializedTestCtx([]byte("abcde"), 2, 1)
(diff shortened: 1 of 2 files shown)


alex-aizman:
I see... previously:

```
pass 1: source => data-slice checksums
pass 2: source => Reed-Solomon encoding
```

with your patch:
```
single pass: source => TeeReader => (data checksums + Reed–Solomon encoding)
```

ec/putjogger_test.go
```diff
@@ -0,0 +1,153 @@
...
+	writers, _ := newParityTestWriters(ctx)
+
+	if err := finalizeSlices(ctx, writers, cksumType); err == nil {
+		t.Fatal("expected unequal data slice lengths to fail encoding")
+	}
+	for i, sl := range ctx.slices {
+		if sl.cksum != nil {
+			t.Fatalf("slice %d published a checksum after failed encoding", i)
+		}
+	}
```
alex-aizman, line 106: This covers the read-error path. For symmetry, it'd be great to have a write-error failure case when there are multiple (concurrent) parity writers, and one of them (randomly) returns an error. Encoding must fail.

alex-aizman:
Separate from the review, this result could be worth a brief technical post or a blog. There's a clean before/after story, and the roughly 2x result at 512 MiB D4/P2 maybe deserves more visibility than a commit message...

ec/putjogger_test.go
ferhatelmas: https://github.com/NVIDIA/aistore/pull/359

ferhatelmas:
> worth a brief technical post or a blog

Sounds good, I’d be happy to write it up. Did you have a particular venue or format in mind?

alex-aizman:
I generally think it's a good idea to publish your results. That's all there is to it.

ferhatelmas:
I see, thank you for the encouragement. I will do and let you know when it's up.

PR #359, opened 2026-08-26
ec: test concurrent parity write failures

ferhatelmas:
* inject a write error into each parity writer in D2 P2 encoding
* verify the failing stream is reported and no checksums

related to #358

### DIFF
--- a/ec/putjogger_test.go
+++ b/ec/putjogger_test.go
@@ -10,9 +10,15 @@ import (
 	"io"
 	"testing"
 
+	"github.com/klauspost/reedsolomon"
+
 	"github.com/NVIDIA/aistore/cmn/cos"
 	"github.com/NVIDIA/aistore/core"
 )
+
+type errorWriter struct{ err error }
+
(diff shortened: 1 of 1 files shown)


ec/putjogger_test.go
```diff
@@ -107,6 +113,43 @@ func TestFinalizeSlicesDoesNotPublishPartialChecksums(t *testing.T) {
...
+func TestFinalizeSlicesFailsOnParityWriteError(t *testing.T) {
+	const cksumType = cos.ChecksumCesXxh
+
+	errInjected := errors.New("injected parity write error")
+	tests := []struct {
+		name      string
+		failIndex int
+	}{
+		{name: "first parity writer", failIndex: 0},
+		{name: "second parity writer", failIndex: 1},
```
alex-aizman, line 125: something like:

```
+		{name: "all parity writers", failIndices: []int{0, 1}},
```

would be good to have

ferhatelmas: https://github.com/NVIDIA/aistore/pull/365

PR #360, opened 2026-08-26
auth: use stdlib for pbkdf2 instead of x/crypto

ferhatelmas:
This is a modernization refactoring, std has the implementation since Go 1.24. 

1. Implementation is identical, x/crypto already [delegates to std](https://cs.opensource.google/go/x/crypto/+/refs/tags/v0.55.0:pbkdf2/pbkdf2.go;l=23).
2. This version is safer and reports errors instead of panicing.
3. FIPS 140 enforcement, hash algorithm suitability, min salt/key sizes, etc.

### DIFF
--- a/cmd/authn/signing/rsa_mgr.go
+++ b/cmd/authn/signing/rsa_mgr.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"crypto/aes"
 	"crypto/cipher"
+	"crypto/pbkdf2"
 	"crypto/rand"
 	"crypto/rsa"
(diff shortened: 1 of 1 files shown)


Nahemah1022:
Hi @ferhatelmas, thanks for opening the PR! Could you update the commit message body to include a bit more context behind this change (e.g., whether it's an optimization or refactoring)? Explain the reasoning for this change

ferhatelmas:
@Nahemah1022 thanks for the nudge, updated the description

aaronnw (approved):

PR #361, opened 2026-09-01
fix: use a valid Go download URL in the dev Docker images

zejj42:
The dev and `test-in-docker` images set `GOLANG_VERSION="1.26"`, which builds `https://go.dev/dl/go1.26.linux-amd64.tar.gz` — a 404 (Go tarballs require the patch version). `curl` then saves the error page and `tar` fails, breaking the image build.

Fix: use `"1.26.0"`, which is a real tarball and matches the `go 1.26` directive in `go.mod`.

```
go1.26.linux-amd64.tar.gz   -> 404
go1.26.0.linux-amd64.tar.gz -> 200
```

Introduced in the "transition to Go 1.26" change; `deploy/ci/build.sh` is unaffected since it resolves the patch version dynamically.

### DIFF
--- a/deploy/dev/docker/Dockerfile
+++ b/deploy/dev/docker/Dockerfile
@@ -26,7 +26,7 @@ RUN wget https://bootstrap.pypa.io/get-pip.py &&\
   pip install awscli s3cmd
 
 # Setting ENV variables
-ENV GOLANG_VERSION="1.26"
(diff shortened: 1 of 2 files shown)


alex-aizman:
we don't want 1.26.0 - we want the latest

today, it'd be 1.26.7

must be scripted - closing

PR #368, opened 2026-09-16
fix: Downloader queue full nil pointer

frankisawesome:
Very self explanatory bug & fix

We previously had a code path that treats all >400 resp codes from the downloader as "error"

But upstream correctly returns no error on a 429, causing a nil pointer deref and the target to crash.

A unit test for the downloader perhaps fits better but this repo seems to prefer a central set of integration test so I've included a regression test there.


alex-aizman:
general guideline: squash two commits

must be a single commit w/ itemized changes in the commit message and sign-off

ais/test/downloader_test.go
```diff
@@ -1399,6 +1401,73 @@ func TestDownloadJobConcurrency(t *testing.T) {
...
+
+	t.Cleanup(func() {
+		for _, id := range ids {
+			if err := api.AbortDownload(baseParams, id); err != nil {
+				t.Errorf("abort %s: %v", id, err)
+			}
+		}
+		for _, id := range ids {
+			waitForDownload(t, id, 30*time.Second)
+		}
```
alex-aizman: This cannot clean up the rejected submission: its ID is never appended to `ids`, while `Xact.Download` stores the job before returning 429.

ais/test/downloader_test.go
```diff
@@ -1456,16 +1468,7 @@ func TestDownloadJobDownloaderQueueFullShouldReturnGracefully(t *testing.T) {
...
-		for _, id := range ids {
-			if err := api.AbortDownload(baseParams, id); err != nil {
-				t.Errorf("abort %s: %v", id, err)
-			}
-		}
-		for _, id := range ids {
-			waitForDownload(t, id, 30*time.Second)
-		}
-	})
+	tassert.Fatalf(t, sawErr, "expected err when downloader queue is full")
```
alex-aizman: wrong assertion - jobs may finish early

ais/test/downloader_test.go
frankisawesome: Outdated

ais/test/downloader_test.go
frankisawesome: Addressed - there's a probably a more deterministic way to do this but for this particular issue I think piggybacking on existing test harness for a quick fix is ok.

frankisawesome:
Thanks for the quick review @alex-aizman addressed 🙏

alex-aizman:
quick review:
* drop the integration test. The 429 doesn't depend on submission count.
* the ais/tgtdl.go fix is a wrong layer
* I'll push a small fix that you can maybe take over and keep going with it

alex-aizman:
this branch:

* https://github.com/NVIDIA/aistore/tree/dload-nilptr

Feel free to take it over - make a PR out of it, add an integration test, and maybe even fix a second and separate bug..

frankisawesome:
I did consider changing it there but figured it's also semantically ok to say 429s are not application errors and don't pass a explicit error. Also happy with that direction @alex-aizman

PR #371, opened 2026-09-18
downloader: discard rejected jobs and preserve live jobs during housekeeping

frankisawesome:
downloader: discard rejected jobs and preserve live jobs during housekeeping

- Remove the in-memory job record when the dispatcher handoff times out.
- Expire only finished jobs so housekeeping retains running and aborting jobs.
- Test rejected and accepted handoffs and housekeeping retention deterministically.

Signed-off-by: Frank Li <b439988l@gmail.com>


alex-aizman:
Hi Frank

I just pushed this commit with a little-bit of conflict resolution and my tiny follow-up: ad5aaa1ea64801089908a5220d3de4bcae7b0386

Closing. Thanks!

